PR #20, opened 2017-09-14
[SDK-72]: Update tests

echarrod:


test/Yoti.Auth.Tests/YotiAttributeValue_Tests.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            string jpegData = "jpegData";
+
+            YotiAttributeValue yotiAttributeValue = new YotiAttributeValue(
+                TypeEnum.Jpeg,
+                jpegData);
+        }
+
+        [TestMethod]
+        public void YotiAttributeValue_PngData_DoesntThrowException()
```
echarrod: I realise the elements that this test covers, as well as the test below, are covered by the most recent commit, so we can remove them.

PR #21, opened 2017-09-18
[SDK-72]: Cleaned up YotiClientEngine

echarrod:
Cleaned up YotiClientEngine - moved relevant parts to CrypoEngine.cs, and broke up very lengthy method into a set of different methods.


vassyz:
@echarrod Are these changes covered by tests?

echarrod:
Aside from the test I mentioned earlier - where we need to add the extra pieces of profile information, the rest of these changes are covered by tests. I can look at covering those areas next.

PR #22, opened 2017-09-19
[SDK-91]: Modified assemblies for new NuGet package

echarrod:
After renaming the solution and project names, it didn't seem right to have both being called Yoti. After researching traditional naming structures, it seemed that Company.ProductName.Project is the best way to go, so I have kept the Project as Yoti.Auth. However, I have changed the PackageId to be Yoti, so that we can reserve this on NuGet, and the user types 
`Install-Package Yoti` to install.

I have changed the version to be 2.0.0. Technically, it could have been 1.0.0, as it's a different package name, but I thought this might be confusing, given the choice between Yoti.Auth 1.1.0, and Yoti 1.0.0.

I removed some of the references in Yoti.Auth.Owin.nuspec, as these are not required. 
I have tested these changes with a local NuGet package repository with the Example and IdentityExample projects. Once these changes have been approved, I will publish the package, and update the Example projects to reference this new package.

### DIFF
--- a/src/Yoti.Auth.Owin.nuspec
+++ b/src/Yoti.Auth.Owin.nuspec
@@ -1,53 +1,39 @@
 <?xml version="1.0"?>
 <package >
   <metadata>
-    <id>Yoti.Auth.Owin</id>
-    <version>1.0.1</version>
-    <authors>Ben Harpin</authors>
-    <owners>Ben Harpin</owners>
+    <id>Yoti.Owin</id>
+    <version>2.0.0</version>
+    <authors>Yoti</authors>
+    <owners>Yoti</owners>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
-    <description>Yoti Owin .NET SDK</description>
-    <copyright>Copyright 2016</copyright>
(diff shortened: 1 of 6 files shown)


src/Yoti.Auth.Owin.nuspec
```diff
@@ -1,53 +1,39 @@
...
-    <authors>Ben Harpin</authors>
-    <owners>Ben Harpin</owners>
+    <id>Yoti.Owin</id>
+    <version>2.0.0</version>
+    <authors>Yoti</authors>
+    <owners>Yoti</owners>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
-    <description>Yoti Owin .NET SDK</description>
-    <copyright>Copyright 2016</copyright>
+    <description>Yoti Owin .NET SDK. Yoti is an identity checking system that allows organisations to verify who people are.</description>
```
vassyz: Where did you get this text?

echarrod: https://www.yoti.com/developers/documentation

PR #26, opened 2017-09-21
[SDK-95]: Fixed bug where Example project was crashing when no photo requested

echarrod:
Example project was crashing when photo was not requested (and consequently supplied) by the application. Also found other bug in the process - if the application changes the details requested of the user, this new information is never added to the user profile.

### DIFF
--- a/src/Example/Controllers/HomeController.cs
+++ b/src/Example/Controllers/HomeController.cs
@@ -15,7 +15,10 @@ namespace Example.Controllers
         {
             var user = GetUser();
 
-            ViewBag.Photo = "data:image/jpeg;base64," + Convert.ToBase64String(user.Photo);
+            if (user.Photo == null)
+                ViewBag.Message = "No photo provided, change the application settings to request a photo from the user for this demo";
+            else
+                ViewBag.Photo = "data:image/jpeg;base64," + Convert.ToBase64String(user.Photo);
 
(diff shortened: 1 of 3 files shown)


src/Example/Controllers/HomeController.cs
```diff
@@ -15,7 +15,10 @@ public ActionResult Index()
         {
             var user = GetUser();
 
-            ViewBag.Photo = "data:image/jpeg;base64," + Convert.ToBase64String(user.Photo);
+            if (user.Photo == null)
+                ViewBag.Message = "No photo provided, change the application settings to request a photo from the user for this demo";
+            else
+                ViewBag.Photo = "data:image/jpeg;base64," + Convert.ToBase64String(user.Photo);
```
vassyz: Shouldn't this happen in the SDK?

echarrod: This code is just choosing what to do with the user profile information retrieved. It is assuming that the user is always sharing their photo. This is fine if this is always specified by the application, but this is just to alert anyone as to why the photo isn't there, as before it was just crashing.

echarrod: As discussed, we'll look at doing the conversion to base64 string in the SDK, so this conversion above doesn't have to be done by every consumer of the API

PR #35, opened 2017-10-19
[SDK-137]: Added Base64Data Helper

echarrod:
Also moved .nuspec objects to the .csproj (and removed .nuspecs), so we don't have to maintain both.

I've already tested that this works on the Example Project via a local NuGet package. Once these changes have been approved I'll publish the package, and then test the online package with the Example project before committing those changes.


src/Example/Models/User.cs
```diff
@@ -9,6 +9,7 @@ public User()
         public int Id { get; set; }
         public string YotiId { get; set; }
         public byte[] Photo { get; set; }
+        public string Base64Photo { get; set; }
```
echarrod: This should have been in the next commit, but I'll leave it there for now, as it doesn't break anything

echarrod:
Last commit includes the URI Media Type, and renamed the variable to Base64URI

vassyz:
Are you planning on mentioning the helper method in the Readme in the next commit?

echarrod:
The README doesn't currently mention the attributes on the User Profile - should it? 

The yoti-web-documentation (in the branch to be merged) lists the attributes like this: 

```csharp
string userId = yotiProfile.Id;
Image Selfie = yotiProfile.Selfie;
string GivenNames = yotiProfile.GivenNames;
string FamilyName = yotiProfile.FamilyName;
string MobileNumber = yotiProfile.MobileNumber;
string EmailAddress = yotiProfile.EmailAddress;
DateTime? DateOfBirth = yotiProfile.DateOfBirth;
string Address = yotiProfile.Address;
string Gender = yotiProfile.Gender;
string Nationality = yotiProfile.Nationality;
```

Do you think we should put something like this in the README too? 

So currently it just shows the Selfie property, I wouldn't have thought we'd need to explicitly point out this new helper method, would you agree?

vassyz:
It makes sense for those attributes and the helper to appear in the Readme. It would probably be worth updating the documentation PR with base64URI method. I'll look into merging that sometime this week.

echarrod:
Added both to the README in the latest commit, I'll do the same in the Yoti-web-documentation if you approve of these changes

PR #36, opened 2017-10-23
[SDK-137]: Updated Example project 

echarrod:
Updated Example project to use latest NuGet package, and changed to display the image the `Base64URI` helper rather than appending the URI media type


vassyz:
Would it be too difficult to add to the example a "Download the selfie file" link that uses the selfie image value?

echarrod:
Added 'Download Image' link in the last commit, displayed below the image

README.md
```diff
@@ -72,6 +72,12 @@ Where:
 * `sdkId` is the sdk identifier generated by Yoti Dashboard when you create (and then publish) your app. _Note this is not your Application Identifier which is needed by your clientside code_
 * `path/to/your-application-pem-file.pem` - When you create your app on Yoti Dashboard, your browser generates a .pem file. This is the path to that file.
 
+You'll need to add the `Callback URL` of your site to the integration settings section on your [Yoti Applications page](https://www.yoti.com/dashboard/applications). This is where the user will be redirected to upon successful authentication. For the Example project, you should append `/account/connect` to the end of this URL.
+
+For debugging purposes, you can use [ngrok](https://ngrok.com/) to create a secure tunnel to localhost by running the following command, and adding the https address generated to your callback URl:
```
vassyz: I'd added ngrok to the Ruby SDK Readme before the Dashboard allowed localhost URLs. I will remove this from the Readme and I don't think it should be mentioned here as well. This will also become less relevant once we have a sandbox app.

README.md
```diff
@@ -72,6 +72,12 @@ Where:
 * `sdkId` is the sdk identifier generated by Yoti Dashboard when you create (and then publish) your app. _Note this is not your Application Identifier which is needed by your clientside code_
 * `path/to/your-application-pem-file.pem` - When you create your app on Yoti Dashboard, your browser generates a .pem file. This is the path to that file.
 
+You'll need to add the `Callback URL` of your site to the integration settings section on your [Yoti Applications page](https://www.yoti.com/dashboard/applications). This is where the user will be redirected to upon successful authentication. For the Example project, you should append `/account/connect` to the end of this URL.
```
vassyz: Can you link `Example` to the example folder. Can we also mention this somewhere where people running the example are likely to look?

echarrod: Done

README.md
echarrod: Removed

vassyz:
Looks good 👍

PR #38, opened 2018-01-30
[SDK-246]: Signed requests

echarrod:
Initial commit for signed requests. I think there is further opportunity for the extraction of functionality from YotIClientEngine, in a similar way to how Java is done, so that this class isn't doing so much, but thought I'd get some feedback on this part first. Tested with Example project and the standard GET request GetActivityDetails is working correctly

### DIFF
--- a/src/Yoti.Auth/Conversion.cs
+++ b/src/Yoti.Auth/Conversion.cs
@@ -19,7 +19,7 @@ namespace Yoti.Auth
         /// UrlSafe Base64 uses '-' instead of '+', and '_' instead of '/' so it
         /// can be passed as a url parameter without extra encoding.
         /// </summary>
-        public static string BytesToUrlsafeBase64(byte[] bytes)
+        public static string BytesToUrlSafeBase64(byte[] bytes)
         {
             var base64 = BytesToBase64(bytes);
 
--- a/src/Yoti.Auth/YotiClientEngine.cs
+++ b/src/Yoti.Auth/YotiClientEngine.cs
@@ -48,7 +48,7 @@ namespace Yoti.Auth
             byte[] httpContent = null;
             HttpMethod httpMethod = HttpMethod.Get;
 
-            string endpoint = GetEndpoint(httpMethod, path, token, nonce, timestamp, sdkId, httpContent);
+            string endpoint = GetEndpoint(httpMethod, path, token, nonce, timestamp, sdkId);
 
             Dictionary<string, string> headers = CreateHeaders(keyPair, httpMethod, endpoint);
 
@@ -56,7 +56,8 @@ namespace Yoti.Auth
                 new HttpClient(),
                 HttpMethod.Get,
                 new Uri(_apiUrl + endpoint),
-                headers);
+                headers,
+                httpContent);
 
             if (response.Success)
             {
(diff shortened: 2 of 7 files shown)


src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -244,14 +256,25 @@ private static void HandleOtherAttributes(YotiUserProfile profile, AttrpubapiV1.
...
         }
 
-        private string GetEndpoint(string token, string nonce, string timestamp, string sdkId)
+        private string GetEndpoint(HttpMethod httpMethod, string path, string token, string nonce, string timestamp, string sdkId, byte[] payload)
         {
-            return string.Format("/profile/{0}?nonce={1}&timestamp={2}&appId={3}", token, nonce, timestamp, sdkId);
+            string endpoint = string.Format("/{0}/{1}?nonce={2}&timestamp={3}&appId={4}", path, token, nonce, timestamp, sdkId);
+            if (payload != null && payload.Length > 0)
+            {
+                string urlSafePayloadString = Conversion.BytesToUrlSafeBase64(payload);
```
qzagarese: It should be base64encoded, not base64URLencoded

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -244,14 +256,25 @@ private static void HandleOtherAttributes(YotiUserProfile profile, AttrpubapiV1.
...
+                endpoint += string.Format("&payload={0}", urlSafePayloadString);
```
qzagarese: please remove "&payload", otherwise it will not work. It should be just "&" + payload_value

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -244,14 +256,25 @@ private static void HandleOtherAttributes(YotiUserProfile profile, AttrpubapiV1.
...
+            }
+
+            return endpoint;
```
qzagarese: If this string is actually used as the url for the request, it will not work (although it accidentally works for GET requests).
This string should be used only as the input for the signed digest that you set as value of the "X-Yoti-Auth-Digest" header.

The url should contain only what you have up to line 261

src/Yoti.Auth/YotiClientEngine.cs
echarrod: Ok, I did it in a similar way to Moussa did for PHP, so does he have to change the PHP one too? - https://github.com/getyoti/yoti-php-sdk/pull/18/files#diff-f97f8bdde6e955330461dafd9d7be9fcR122

qzagarese: Correct! Thanks for spotting this!

src/Yoti.Auth/YotiClientEngine.cs
echarrod: So will the URL be handled correctly if the payload contains "`+`" or "`/`"s?

src/Yoti.Auth/YotiClientEngine.cs
echarrod: I've sent Moussa a message so he is aware of the necessary change

src/Yoti.Auth/YotiClientEngine.cs
echarrod: After seeing your below comment - this makes sense. So it's not being added to the URL, and so doesn't need to be URL safe

src/Yoti.Auth/YotiClientEngine.cs
echarrod: Ok sure, and the payload/content should be added as request.Content in [HttpRequester](https://github.com/getyoti/yoti-dotnet-sdk/pull/38/files#diff-26f682868312f987ad359165eadb85aa)?

qzagarese: Yes, as it was supposed to be there (not if it is a GET obviously)

src/Yoti.Auth/Conversion.cs
```diff
@@ -19,7 +19,7 @@ public static string BytesToBase64(byte[] bytes)
         /// UrlSafe Base64 uses '-' instead of '+', and '_' instead of '/' so it
         /// can be passed as a url parameter without extra encoding.
         /// </summary>
-        public static string BytesToUrlsafeBase64(byte[] bytes)
+        public static string BytesToUrlSafeBase64(byte[] bytes)
```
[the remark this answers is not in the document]
echarrod: There is another method in this class which is called `BytesToBase64`. `BytesToUrlSafeBase64` converts `+` to `-` and `/` to `_`, so that it can be used in a URL, although it's not being called from anywhere currently, but the reverse method `UrlSafeBase64ToBytes` is.

PR #39, opened 2018-02-08
[SDK-258]: Profile attributes

echarrod:
Adding the new attributes requires the DB to be updated too. 

To avoid doing passing attributes through the ViewBag, I changed the Profile page to be handled by the Account Controller, so that we can easily access the attributes through the model.


echarrod:
Need to add the new attributes to the README too, will add those to this PR

src/Example/Controllers/AccountController.cs
```diff
@@ -79,6 +86,20 @@ public ActionResult Connect(string token)
             }
         }
 
+        private static void UpdateAttributesIfPresent(YotiUserProfile yotiProfile, User user)
```
qzagarese: Is this an attempt to align the host application user with the values from the profile using reflection?

echarrod: Since the .NET example is saving to a database, this method saves the values retrieved from the profile to an object which is then saved to the database

PR #40, opened 2018-02-20
[SDK-272]: Set default HTTP security protocol to be TLS 1.2

echarrod:
Going forwards, we may be asking clients to use .NET 4.6+, but in the meantime, this fix will ensure that they use TLS 1.2 to match Connect. 

System.Net.ServicePointManager is not available with .NET Core, since it is totally decoupled from System.Net, 
This is fine though as for .NET 4.6+ TLS 1.2 is supported by default.


echarrod:
Branched at the wrong point with this change - I made a new PR which branches from the point of the latest release - 2.2.0 https://github.com/getyoti/yoti-dotnet-sdk/compare/TLS-specifying?expand=1

PR #41, opened 2018-02-20
[SDK-272]: Set default HTTP security protocol to be TLS 1.2

echarrod:
Going forwards, we may be asking clients to use .NET 4.6+, but in the meantime, this fix will ensure that they use TLS 1.2 to match Connect.

System.Net.ServicePointManager is not available with .NET Core, since it is totally decoupled from System.Net,
This is fine though as for .NET 4.6+ TLS 1.2 is supported by default.


echarrod:
Merged master into this branch so that the merge into master can occur once changes are approved. 
v2.2.1 with this fix will be released from just the first check in though, not including the merge.

PR #44, opened 2018-03-21
[SDK-281]: IsAgeVerified function

echarrod:
- Also changed Example project to use .env variables (or the .NET equivalent - secrets.config)


src/Yoti.Auth.Owin/YotiAuthenticationExtensions.cs
```diff
@@ -141,12 +141,28 @@ public static YotiUserProfile GetYotiProfile(this ExternalLoginInfo loginInfo)
                         break;
 
                     default:
-                        YotiAttributeValue.TypeEnum valueType;
-                        if (Enum.TryParse(claim.ValueType, out valueType))
+                        if (claim.Type.StartsWith(Constants.AttributeAgeOver)
```
vassyz: Can't the duplication from `src/Yoti.Auth/Activity.cs` be avoided? Maybe the behaviour to be moved into a method `parseAgeVerified`.

echarrod: I did think of doing that, but we won't need this whole project once we have the .NET core example working. .NET core will use the same code in Activity.cs

vassyz: OK, makes sense 👍

PR #50, opened 2018-07-05
SDK-375: Added extra logging for .NET Core example project

echarrod:


echarrod:
We have `_logger.LogInformation`, `_logger.LogWarning` and `_logger.LogError used here`, is that what you mean?

PR #53, opened 2018-08-15
Fixed Documentation Link in README.md

CharlesVu:
## Purpose
Fixed Documentation Link

## Approach
External links needs to be prefixed with `https://` otherwise Github will prefix it with 
`https://github.com/getyoti/yoti-dotnet-sdk/blob/master/` as it thinks it is an internal link.

## Scope of changes
README.md

## Checklist
- [ ] Check that the link is redirecting correctly to `https://www.yoti.com/developers/documentation/?csharp` an not `https://github.com/getyoti/yoti-dotnet-sdk/blob/master/www.yoti.com/developers/documentation/?csharp`

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,4 +273,4 @@ Please provide the following to get you up and working as quickly as possible:
 
 Once we have answered your question we may contact you again to discuss Yoti products and services. If you’d prefer us not to do this, please let us know when you e-mail.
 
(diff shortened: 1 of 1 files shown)


echarrod:
Looks good, not sure why `https` was in the description but not in the actual link

PR #54, opened 2018-10-05
[SDK-209]: Token & Tidying up

echarrod:
- Change token -> oneTimeUseToken
- General tidying up:
  - Correct comments
  - Correct namespaces
  - Refactor tests


src/Yoti.Auth/CryptoEngine.cs
```diff
@@ -26,20 +26,20 @@ public static AsymmetricCipherKeyPair LoadRsaKey(StreamReader keyStream)
 
         public static byte[] DecipherAes(byte[] key, byte[] iv, byte[] cipherBytes)
         {
-            KeyParameter keyParam = new KeyParameter(key);
-            ParametersWithIV keyParamWithIv = new ParametersWithIV(keyParam, iv);
+            var keyParam = new KeyParameter(key);
```
bucky-boy: Why use _var_ everywhere if you know what the expected types are?

src/Yoti.Auth/CryptoEngine.cs
```diff
@@ -58,7 +58,7 @@ public static byte[] DecryptRsa(byte[] cipherBytes, AsymmetricCipherKeyPair keyp
         public static byte[] SignDigest(byte[] digestBytes, AsymmetricCipherKeyPair keypair)
         {
             // create a signature from the digest using SHA256 hashing with RSA
-            var signer = SignerUtilities.GetSigner(DigestAlgorithm);
+            ISigner signer = SignerUtilities.GetSigner(DigestAlgorithm);
```
bucky-boy: But here we've changed it the other way

src/Yoti.Auth/DataObjects/ProfileDO.cs
```diff
@@ -2,8 +2,8 @@
 {
     internal class ProfileDO
     {
-        public string session_data { get; set; }
+        public string SessionData { get; set; }
```
bucky-boy: We want these object to be immutable when returned to SDK clients.  We need to look at the use of _Set_ property.  Maybe use _internal_ or something, rather than _public_, if we really need to use _Set_ at all

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -23,33 +23,23 @@ public YotiClientEngine(IHttpRequester httpRequester)
 #endif
         }
 
-        /// <summary>
```
bucky-boy: Why is the documentation going?

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -118,23 +98,23 @@ public async Task<AmlResult> PerformAmlCheckAsync(string appId, AsymmetricCipher
...
+            Dictionary<string, string> headers = CreateHeaders(keyPair, httpMethod, endpoint, httpContent);
 
             AmlResult result = await Task.Run(async () => await new RemoteAmlService()
                 .PerformCheck(_httpRequester, amlProfile, headers, apiUrl, endpoint, httpContent));
 
             return result;
         }
 
-        private static Dictionary<string, string> CreateHeaders(AsymmetricCipherKeyPair keyPair, HttpMethod httpMethod, string endpoint, byte[] httpContent, string contentType)
+        public static Dictionary<string, string> CreateHeaders(AsymmetricCipherKeyPair keyPair, HttpMethod httpMethod, string endpoint, byte[] httpContent)
```
bucky-boy: Not sure we want to expose this as _public_...?

test/Yoti.Auth.Tests/ActivityTests.cs
```diff
@@ -125,11 +125,11 @@ public void Activity_AddAttributesToProfile_FullName()
         [TestMethod]
         public void Activity_AddAttributesToProfile_GivenNames()
         {
-            string value = "GivenNamesValue";
+            const string value = "GivenNamesValue";
 
             var attribute = new AttrpubapiV1.Attribute
             {
-                Name = _givenNamesAttribute,
+                Name = GivenNamesAttribute,
```
bucky-boy: We don't need to define a different test value for every test.  We only care that it's a String.  We set the String value, then prove that we get it back when we ask for it.

We can just define _"someStringValue"_ or similar, then reuse it.

test/Yoti.Auth.Tests/ActivityTests.cs
```diff
@@ -10,50 +10,50 @@
...
+        private const string SubdistrictJson = "subdistrict";
+        private const string DistrictJson = "district";
+        private const string PostOfficeJson = "post_office";
+
+        private const string YotiAdminVerifierType = "YOTI_ADMIN";
+        private const string PassportSourceType = "PASSPORT";
+        private const string DrivingLicenseSourceType = "DRIVING_LICENCE";
+
+        private const string GivenNamesAttribute = "given_names";
+        private const string FamilyNameAttribute = "family_name";
```
bucky-boy: If we only use these once, we might as well just inline them

src/Yoti.Auth/CryptoEngine.cs
bucky-boy: Ok.  The advantage of _var_ seems to be it allows for anonymous types.
https://stackoverflow.com/questions/3425966/what-advantages-does-using-var-have-over-the-explicit-type-in-c

src/Yoti.Auth/CryptoEngine.cs
echarrod: So it's only if the explicit type `ISigner` is used on the right hand side, that we use var. Although you could guess that it would be of type `ISigner`, it isn't explicit

src/Yoti.Auth/DataObjects/ProfileDO.cs
echarrod: Since the `ProfileDO` class is internal, everything within it can at most be internal (and not public). 

The convention is to use public for everything within a class, as this just signifies to observe the top level accessibility defined (which is `internal` here).

src/Yoti.Auth/YotiClientEngine.cs
echarrod: We currently only have documentation for the client-facing methods. Since the methods here all come through `YotiClient`, as this is the client-facing code, then we don't need it in `YotiClientEngine` too.

src/Yoti.Auth/YotiClientEngine.cs
echarrod: Hmm yes not sure why that happened, will change

test/Yoti.Auth.Tests/ActivityTests.cs
echarrod: We use GivenNames, Selfie, AgeOver18, AgeUnder18, PostalAddress, StructuredPostalAddress & DateOfBirth in multiple places. I thought it was quite neat having them all together, but do you think it would be better to inline the ones that aren't?

test/Yoti.Auth.Tests/ActivityTests.cs
echarrod: Done 👍

test/Yoti.Auth.Tests/ActivityTests.cs
bucky-boy: Yes.  These tests are bulging.  Or if we really want them defined as constants, pull them out to another constants class - or maybe just use the constants defined in the SDK already

src/Yoti.Auth/CryptoEngine.cs
bucky-boy: I'm not keen on this.  If we don't have a reason _not_ to specify an explicit type, I don't see why we wouldn't want to.  Using _var_ seems to imply that the variable could be something other than _KeyParameter_, but that's not our intention

echarrod: I would agree if it was something like `var i = "42"`, where the compiler just converts it to a string. 

But here, using var when the type is declared on the RHS will always use that type, It's really just a coding style.

src/Yoti.Auth/Activity.cs
```diff
@@ -7,15 +7,16 @@
...
 using Yoti.Auth.DataObjects;
 using static Yoti.Auth.YotiAttributeValue;
 
 namespace Yoti.Auth
 {
     internal class Activity
     {
-        private YotiUserProfile _yotiUserProfile;
-        private YotiProfile _yotiProfile;
+        private readonly YotiUserProfile _yotiUserProfile;
```
sidibos: Some comments here will be good to know which one is the old profile.

echarrod: Visual Studio underlines the deprecated methods, and shows this message when you hover over `YotiUserProfile`: image

This will happen any place that YotiUserProfile, or the fields within it are used. Do you think it's still necessary to specify this is old, given this?

PR #55, opened 2018-10-18
[SDK-554] Adding ReceiptID to ActivityDetails

echarrod:
- Added ReceiptID to ActivityDetails
- Updated ActivityDetails items to be privately set
- Corrected package locations for 45Example project
- Changed CoreExample project to use inline QR code

### DIFF
--- a/src/Yoti.Auth/ActivityDetail.cs
+++ b/src/Yoti.Auth/ActivityDetail.cs
@@ -32,17 +32,17 @@ namespace Yoti.Auth
         /// The <see cref="YotiUserProfile"/> returned by Yoti if the request was successful.
         /// </summary>
         [Obsolete("Please use Profile instead")]
-        public YotiUserProfile UserProfile { get; private set; }
+        public YotiUserProfile UserProfile { get; set; }
 
         /// <summary>
         /// The <see cref="YotiProfile"/> returned by Yoti if the request was successful.
         /// </summary>
-        public YotiProfile Profile { get; private set; }
+        public YotiProfile Profile { get; set; }
 
         /// <summary>
         /// The outcome status of the request.
         /// </summary>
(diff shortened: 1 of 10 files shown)


src/Yoti.Auth/ActivityDetail.cs
```diff
@@ -12,20 +12,42 @@ public enum ActivityOutcome { Success, ProfileNotFound, Failure, SharingFailure
...
         /// The <see cref="YotiProfile"/> returned by Yoti if the request was successful.
         /// </summary>
-        public YotiProfile Profile { get; set; }
+        public YotiProfile Profile { get; private set; }
 
         /// <summary>
         /// The outcome status of the request.
         /// </summary>
-        public ActivityOutcome Outcome { get; set; }
+        public ActivityOutcome Outcome { get; private set; }
```
bucky-boy: Technically, although no-one should be using this, it is a breaking change isn't it

echarrod: Yes you're right, I'll have to delay making these private until v3.0.0

src/Yoti.Auth/ActivityDetail.cs
```diff
@@ -12,20 +12,42 @@ public enum ActivityOutcome { Success, ProfileNotFound, Failure, SharingFailure
     /// </summary>
     public class ActivityDetails
     {
+        public ActivityDetails(ActivityOutcome activityOutcome)
```
sidibos: Just curious - why do we need two constructors?

echarrod: If the share is successful we use one with 4 args, if it's not successful, we use the other one with just the outcome, this was just a way of making the object immutable after being constructed.

PR #56, opened 2018-10-24
[SDK-554]: ActivityDetails

echarrod:
- Add methods to ActivityDetails, using Java as a template:
  - ApplicationProfile
  - rememberMeId (Id will be removed from profile in v3.0.0)
  - timestamp
  - receiptID

Updated ActivityDetails to throw exception if the share is not successful, with `Outcome` being removed in v3.0.0

`ApplicationProfile` inherits from `Profile`, which has the `getAttributeByName` function. 
`UserProfile` will inherit from this object too in the next release.


test/Yoti.Auth.Tests/TestTools/Profile.cs
```diff
@@ -0,0 +1,24 @@
...
+using static Yoti.Auth.YotiAttributeValue;
+
+namespace Yoti.Auth.Tests.TestTools
+{
+    internal class Profile
+    {
+        public static ApplicationProfile CreateApplicationProfileWithSingleAttribute(byte[] value, string name, TypeEnum type)
+        {
+            var yotiAttributeValue = new YotiAttributeValue(
+                type,
```
bucky-boy: Why all the carriage returns?

src/Yoti.Auth/Activity.cs
```diff
@@ -40,21 +40,54 @@ public ActivityDetails HandleSuccessfulResponse(AsymmetricCipherKeyPair keyPair,
             {
```
bucky-boy: The previous two statements both _return_, so we don't need this to be in an _else_ block

src/Yoti.Auth/Activity.cs
```diff
@@ -40,21 +40,54 @@ public ActivityDetails HandleSuccessfulResponse(AsymmetricCipherKeyPair keyPair,
...
+                var applicationProfileAttributes = ParseApplicationProfileContent(keyPair, parsedResponse);
 
-                _yotiUserProfile.Id = parsedResponse.Receipt.remember_me_id;
-                _yotiProfile.Id = parsedResponse.Receipt.remember_me_id;
-                string receiptId = parsedResponse.Receipt.receipt_id;
+                ApplicationProfile applicationProfile = new ApplicationProfile(applicationProfileAttributes);
+                DateTime? timestamp = null;
 
-                AddAttributesToProfile(attributes);
+                if (DateTime.TryParseExact(parsedResponse.Receipt.timestamp, "yyyy-MM-ddTHH:mm:ssZ", CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTime parsedDate))
```
bucky-boy: 1. Can _timestamp_ be null?
2. What happens if parsing fails?  Does _TryParseExact_ throw an Exception?

src/Yoti.Auth/Profile.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Retrieves an attribute based on its name
+        /// </summary>
+        /// <typeparam name="T">The expected type of the attribute</typeparam>
+        /// <param name="name">The name of the attribute</param>
+        /// <returns></returns>
+        public YotiAttribute<T> GetAttributeByName<T>(string name)
+        {
+            if (_attributes.TryGetValue(name, out YotiAttribute<object> matchingAttribute))
+            {
+                return new YotiAttribute<T>(matchingAttribute);
```
bucky-boy: So, you have to use _new_ to have the correct generic type?

test/Yoti.Auth.Tests/ApplicationProfileTests.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    [TestClass]
+    public class ApplicationProfileTests
+    {
+        private string _value = "value";
+
+        [TestMethod]
+        public void ApplicationProfile_NameAttribute()
+        {
+            ApplicationProfile applicationProfile = TestTools.Profile.CreateApplicationProfileWithSingleAttribute(
```
bucky-boy: Don't we have a mocking framework?

test/Yoti.Auth.Tests/ProfileTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            });
+        }
+
+        [TestMethod]
+        public void ProfileGetAttribute_Image()
+        {
+            string value = "selfie0123456789";
+
+            var yotiAttributeValue = new YotiAttributeValue(TypeEnum.Jpeg, Encoding.UTF8.GetBytes(value));
+            var yotiAttribute = new YotiAttribute<object>(YotiConstants.SelfieAttribute, yotiAttributeValue);
```
bucky-boy: YotiAttribute<Image> ?

test/Yoti.Auth.Tests/ProfileTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            Assert.ThrowsException<InvalidCastException>(() =>
+            {
+                dobAttribute.GetValue();
+            });
+        }
+
+        [TestMethod]
+        public void ProfileGetAttribute_Image()
+        {
+            string value = "selfie0123456789";
```
bucky-boy: If the value is only used once, please inline it

src/Yoti.Auth/Activity.cs
```diff
@@ -40,21 +40,54 @@ public ActivityDetails HandleSuccessfulResponse(AsymmetricCipherKeyPair keyPair,
             {
                 ReceiptDO receipt = parsedResponse.Receipt;
 
-                AttrpubapiV1.AttributeList attributes = CryptoEngine.DecryptCurrentUserReceipt(
-                    parsedResponse.Receipt.wrapped_receipt_key,
-                    parsedResponse.Receipt.other_party_profile_content,
-                    keyPair);
+                ParseOtherPartyProfileContent(keyPair, parsedResponse);
+                var applicationProfileAttributes = ParseApplicationProfileContent(keyPair, parsedResponse);
```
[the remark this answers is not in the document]
echarrod: Good spot, will change

src/Yoti.Auth/Activity.cs
echarrod: Yep good policy, nice to get rid of noise when we can

src/Yoti.Auth/Activity.cs
echarrod: 1. Will add null check
2. [Looks like](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.tryparseexact?view=netframework-4.7.2) it can just throw an `ArgumentException`

test/Yoti.Auth.Tests/ApplicationProfileTests.cs
echarrod: I have created https://lampkicking.atlassian.net/browse/SDK-656 to introduce this

src/Yoti.Auth/ApplicationProfile.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System.Collections.Generic;
+
+namespace Yoti.Auth
+{
+    /// <summary>
+    /// Profile of an application, with convenience methods to access well-known attributes.
+    /// </summary>
+    public class ApplicationProfile : Profile
+    {
+        public ApplicationProfile() : base()
```
bucky-boy: I don't think we want these constructors to be public.
Only we should be able to create instances of the Profiles + the Attributes.

src/Yoti.Auth/Profile.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            _attributes = new Dictionary<string, BaseAttribute>();
+        }
+
+        internal Profile(Dictionary<string, BaseAttribute> attributes)
+        {
+            _attributes = attributes;
+        }
+
+        internal void Add(BaseAttribute value)
```
bucky-boy: I wonder if this can still accept _YotiAttribute<T>_ as a parameter.
Then the fact that it is actually working with _BaseAttribute_ is kept entirely internal, and there is no possibility of adding the wrong kind of _BaseAttribute_ to the Dictionary.

src/Yoti.Auth/ApplicationProfile.cs
echarrod: Yes good spot, will change

src/Yoti.Auth/Profile.cs
echarrod: Yep just wrote a test for that scenario and it works, will check in

src/Yoti.Auth/Activity.cs
```diff
@@ -36,23 +36,57 @@ public ActivityDetails HandleSuccessfulResponse(AsymmetricCipherKeyPair keyPair,
...
+            if (receipt.timestamp != null)
+            {
+                if (DateTime.TryParseExact(receipt.timestamp, "yyyy-MM-ddTHH:mm:ssZ", CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTime parsedDate))
+                {
+                    timestamp = parsedDate;
+                }
+            }
 
-                return new ActivityDetails(_yotiUserProfile, _yotiProfile, receiptId, ActivityOutcome.Success);
+            return new ActivityDetails(parsedResponse.Receipt.remember_me_id, timestamp, _yotiUserProfile, _yotiProfile, applicationProfile, parsedResponse.Receipt.receipt_id, ActivityOutcome.Success);
```
sidibos: This is just a suggestion: Will it be better if pass the whole Receipt Object to ActivityDetails constructor?
As ActivityDetails(Receipt $receipt, ..) as rememberMe_Id, timestamp, and Receipt_id are all from Receipt.
The decision is yours, this is just a suggestion.

src/Yoti.Auth/Activity.cs
```diff
@@ -36,23 +36,57 @@ public ActivityDetails HandleSuccessfulResponse(AsymmetricCipherKeyPair keyPair,
...
+        }
+
+        private void ParseOtherPartyProfileContent(AsymmetricCipherKeyPair keyPair, ProfileDO parsedResponse)
+        {
+            AttrpubapiV1.AttributeList profileAttributes = CryptoEngine.DecryptCurrentUserReceipt(
```
sidibos: You might want to eliminate decryption redundancy for both OtherProfileContent and ProfileContent, by moving this into one place AttributesListConverter::class.
Again this is a suggestion.

src/Yoti.Auth/YotiConstants.cs
```diff
@@ -26,5 +27,10 @@ internal static class YotiConstants
         public const string PostalAddressAttribute = "postal_address";
         public const string StructuredPostalAddressAttribute = "structured_postal_address";
         public const string DateOfBirthAttribute = "date_of_birth";
+
```
sidibos: I would have moved these constants in their relevant classes Profile and ApplicationProfile.
Just a suggestion!

src/Yoti.Auth/Activity.cs
echarrod: That's a good suggestion, but since `ReceiptDO` has an `internal` access modifier and is less accessible than `ActivityDetails`, if we were to pass in the whole receipt data object, we'd have to convert `ReceiptDO` to `public`. I think we should keep it as internal, even though it would make it slightly neater if we passed the whole receipt in.

src/Yoti.Auth/YotiConstants.cs
echarrod: Yes fair enough, now that this has become large it makes sense

src/Yoti.Auth/Activity.cs
echarrod: I've added this to https://lampkicking.atlassian.net/browse/SDK-587 now

PR #57, opened 2018-10-30
[SDK-554]: Update README

echarrod:
- Update README to have:
  - `RememberMeID`
  - `ApplicationProfile` in API Coverage
- Change AttributeConverter to not return base attribute type (and instead the derived class), so we can be more specific


README.md
```diff
@@ -260,6 +259,11 @@ bool onWatchList = amlResult.IsOnWatchList();
     * [X] Postal Address `Address`
     * [X] Gender `Gender`
     * [X] Nationality `Nationality`
+  * [X] ApplicationProfile
+    * [X] Name `Name`
+	* [X] URL `URL`
```
sidibos: These look indented compare to Name.

PR #58, opened 2018-10-30
development -> master

echarrod:
Proposed release notes:

### Added
- To ActivityDetails:
  - ApplicationProfile
    - Name
    - URL
    - Logo
    - Receipt Background Colour
  - RememberMeId (Id will be removed from profile in v3.0.0)
  - Timestamp
  - ReceiptID

### Changed
- ActivityDetails to throw exception if the share is not successful, `Outcome` still functions the same, but will be removed in v3.0.0

### Deprecated
- Profile.Id (replaced by ActivityDetails.RememberMeID)


src/Yoti.Auth/BaseAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace Yoti.Auth
+{
+    /// <summary>
+    /// A class to represent a base Yoti attribute, without any generics. A Yoti attribute
+    /// consists of the attribute name, an associated <see cref="YotiAttributeValue"/>,
+    /// and a list of <see cref="Anchor"/>s from Yoti. It may hold one or more anchors,
+    /// which specify how an attribute has been provided and how it has been verified
+    /// within the Yoti platform.
+    /// </summary>
+    public class BaseAttribute
```
bucky-boy: 1. Should it be public?
2. Should it be abstract?

src/Yoti.Auth/BaseAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        private readonly string _name;
+        private List<Anchor> _anchors;
+
+        public BaseAttribute(string name, YotiAttributeValue value)
```
bucky-boy: Again, does this need to be public?

src/Yoti.Auth/YotiAttribute.cs
```diff
@@ -1,42 +1,37 @@
...
     /// <summary>
-    /// A class to represent a Yoti attribute. A Yoti attribute consists of the attribute name, an associated
+    /// A class to represent a Yoti attribute, extending <see cref="BaseAttribute"/>.
+    /// A Yoti attribute consists of the attribute name, an associated
     /// <see cref="YotiAttributeValue"/>, and a list of <see cref="Anchor"/>s from Yoti.
     /// It may hold one or more anchors, which specify how an attribute has been provided
     /// and how it has been verified within the Yoti platform.
     /// </summary>
-    public class YotiAttribute<T>
+    public class YotiAttribute<T> : BaseAttribute
```
bucky-boy: I guess in here it's all public to avoid a breaking change...?

src/Yoti.Auth/BaseAttribute.cs
echarrod: 1. Since the derived class YotiAttribute is public, we can't change BaseAttribute without also changing YotiAttribute, as it would be less accessible.
1. Yes, good idea

src/Yoti.Auth/BaseAttribute.cs
echarrod: Will change this to be `private protected`

src/Yoti.Auth/YotiAttribute.cs
echarrod: As this is the return type of the attributes which the clients retrieve, we need to make it public

echarrod:
Reverted last commit, which will be included in v3.0.0

src/Yoti.Auth/AttributeConverter.cs
```diff
@@ -47,6 +50,56 @@ public static YotiAttribute<object> ConvertAttribute(AttrpubapiV1.Attribute attr
...
+                        ParseAnchors(attribute));
+
+                case ContentType.Json:
+                    value = new YotiAttributeValue(TypeEnum.Json, attribute.Value.ToByteArray());
+                    return new YotiAttribute<IEnumerable<Dictionary<string, JToken>>>(
+                        attribute.Name,
+                        value,
+                        ParseAnchors(attribute));
+
+                case ContentType.Undefined:
```
sidibos: Maybe you could combine these 2 into default?

echarrod: I guess it might not be as clear that the comment pertains to undefined attributes if we were to couple it with the default case.

PR #59, opened 2018-11-08
[SDK-587]: Profile attributes, tidying up

echarrod:
- Change UserProfile to be a dictionary
- Remove obsolete methods & Owin project


test/Yoti.Auth.Tests/ProfileTests.cs
```diff
@@ -10,16 +10,16 @@ namespace Yoti.Auth.Tests
...
     {
         [TestMethod]
-        public void Profile_GetAttribute()
+        public void Profile_GetAttribute_String()
         {
             var attributeValue = new YotiAttributeValue(TypeEnum.Text, Encoding.UTF8.GetBytes("value"));
-            var initialAttribute = new YotiAttribute<string>(Constants.UserProfile.PhoneNumberAttribute, attributeValue);
+            var initialAttribute = new YotiAttribute<string>(Constants.ApplicationProfile.ApplicationNameAttribute, attributeValue);
 
             ApplicationProfile applicationProfile = TestTools.Profile.CreateApplicationProfileWithSingleAttribute(initialAttribute);
```
bucky-boy: It's not immediately clear which class you are actually testing.
This test is for the base object, correct?  Or is it really for ApplicationProfile?

test/Yoti.Auth.Tests/ProfileTests.cs
```diff
@@ -29,9 +29,25 @@ public void Profile_GetAttribute_Datetime()
...
+            Assert.AreSame(initialAttribute, dobAttribute);
+            DateTime.TryParseExact(value, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTime dob);
+            Assert.AreEqual(dob, dobAttribute.GetValue());
+        }
+
+        [TestMethod]
+        public void Profile_GetAttribute_Datetime_Nullable()
+        {
+            string value = "1980-01-13";
+            var attributeValue = new YotiAttributeValue(TypeEnum.Date, Encoding.UTF8.GetBytes(value));
```
bucky-boy: The _YotiAttributeValue_ wrapper must die!
We can just parse to the appropriate native data type and keep that in the attribute.

src/Yoti.Auth/AttributeConverter.cs
```diff
@@ -65,7 +24,7 @@ public static BaseAttribute ConvertToBaseAttribute(AttrpubapiV1.Attribute attrib
 
                 case ContentType.Date:
                     value = new YotiAttributeValue(TypeEnum.Date, attribute.Value.ToByteArray());
-                    return new YotiAttribute<DateTime>(
+                    return new YotiAttribute<DateTime?>(
```
bucky-boy: I don't think the DateTime needs to be nullable.
We can parse attribute values up front, as soon as we get the receipt.  If there is an error parsing any given DateTime, we log it, and we don't add that attribute to the profile.

test/Yoti.Auth.Tests/ProfileTests.cs
echarrod: Yes this file is for the base object. I guess it would be clearer if I split it up into `ApplicationProfileTests` and `UserProfileTests`, would you agree?

src/Yoti.Auth/AttributeConverter.cs
echarrod: I've created https://lampkicking.atlassian.net/browse/SDK-679 to implement that

test/Yoti.Auth.Tests/ProfileTests.cs
echarrod: I've added https://lampkicking.atlassian.net/browse/SDK-680 to remove this

src/Yoti.Auth/AttributeConverter.cs
bucky-boy: Ok.  But I still don't see any need to change it to nullable in this PR, even if we do intend to take it out in a couple of days time.

test/Yoti.Auth.Tests/ProfileTests.cs
bucky-boy: Yes, if you are testing two different objects then please have two different tests.  

If you're only testing _ApplicationProfile_ then just call it _ApplicationProfileTest_.

If you're deliberately testing the base class, and you don't really care about _UserProfile_ or _ApplicationProfile_ here, then one approach is to create a simple subclass within your test (say, _TestProfile_) and just test that.

src/Yoti.Auth/AttributeConverter.cs
echarrod: Ah sorry, didn't see what you meant. Changed now

test/Yoti.Auth.Tests/UserProfileTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using static Yoti.Auth.YotiAttributeValue;
+
+namespace Yoti.Auth.Tests
+{
+    [TestClass]
+    public class UserProfileTests
+    {
+        [TestMethod]
+        public void UserProfile_GetAttribute_Datetime()
```
qzagarese: It would be nice to have more descriptive names.
Ideally I would not need to read the body of the test to understand what's actually checked.
I usually use the form `testSomething` or the `shouldSomething`/`somethingShould`, which in this case would be something like `testParsedDateMatchesInitialValue` or `parsedDateShouldMatchInitialValue`

test/Yoti.Auth.Tests/UserProfileTests.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            YotiAttribute<DateTime?> dobAttribute = userProfile.GetAttributeByName<DateTime?>(Constants.UserProfile.DateOfBirthAttribute);
+
+            Assert.AreSame(initialAttribute, dobAttribute);
+            DateTime.TryParseExact(value, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTime dob);
+            Assert.AreEqual(dob, dobAttribute.GetValue());
+        }
+
+        [TestMethod]
+        public void UserProfile_GetAttribute_Datetime_InvalidFormat()
```
qzagarese: same here

test/Yoti.Auth.Tests/UserProfileTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            YotiAttribute<DateTime?> dobAttribute = userProfile.GetAttributeByName<DateTime?>(Constants.UserProfile.DateOfBirthAttribute);
+
+            Assert.ThrowsException<InvalidCastException>(() =>
+            {
+                dobAttribute.GetValue();
+            });
+        }
+
+        [TestMethod]
+        public void UserProfile_AddAttribute()
```
qzagarese: same here

test/Yoti.Auth.Tests/YotiProfileTests.cs
```diff
@@ -0,0 +1,147 @@
...
+
+namespace Yoti.Auth.Tests
+{
+    [TestClass]
+    public class YotiProfileTests
+    {
+        private readonly string _value = "value";
+
+        [TestMethod]
+        public void YotiProfile_SelfieAttribute()
```
qzagarese: same for the tests in this class

test/Yoti.Auth.Tests/UserProfileTests.cs
echarrod: I agree that more descriptive names would be useful so you can see exactly what the behaviour should be from the name along. Additionally, we don't need to prefix the tests with the class (`UserProfile` here), since that can be inferred from the name of the tests.

I would like the test naming to be consistent across the tests though, so if we are to rename them, I would like them all to be renamed that way. I have created a ticket for this - https://lampkicking.atlassian.net/browse/SDK-682

PR #60, opened 2018-11-26
SDK-591 SDK-630 SDK-679: Browser.js, Image Object, Removals

echarrod:
- Remove:
  - YotiAttributeValue
  - GetValueorDefault
  - Extra TypeEnum - now uses Protobuf types
- Move methods to YotiAttribute
- Update browser.js
- Update test NuGet packages - just for improved code coverage reporting
- Changed DateTime attributes to be non-nullable


src/Yoti.Auth/Image.cs
```diff
@@ -1,32 +1,49 @@
...
         {
             get
             {
                 switch (Type)
                 {
-                    case TypeEnum.Jpeg:
-                        return "data:image/jpeg;base64," + Conversion.BytesToBase64(Data);
-                    case TypeEnum.Png:
-                        return "data:image/png;base64," + Conversion.BytesToBase64(Data);
+                    case AttrpubapiV1.ContentType.Jpeg:
```
bucky-boy: If we take the approach of using different subclasses for different image types, then:
1. This _Image_ class does not need to know about _AttrpubapiV1.ContentType_ i.e. it becomes agnostic towards our underlying implementation - it knows nothing about Protobuf (and it shouldn't really)
2. Specific classes will know how to represent themselves - we will not need multiple switch-case statements

src/Yoti.Auth/YotiAttribute.cs
```diff
@@ -13,25 +15,40 @@ namespace Yoti.Auth
...
+            _type = type;
+            _data = data;
+        }
+
+        public byte[] Data()
+        {
+            return _data;
+        }
+
+        public AttrpubapiV1.ContentType Type()
```
bucky-boy: This looks like we are exposing ContentType from the protobuf....  we must hide that away.  No protobuf stuff should ever be exposed.

And I don't think we need to have an additional _Type_ on _Attribute_ at all.  We already have generic types for _Attribute_ that determine the type of the underlying value.

src/Yoti.Auth/Images/Image.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Yoti.Auth.Images
+{
+    public abstract class Image
```
echarrod: Do you think this should be called `BaseImage` or `AbstractImage` so that we don't have as much of a stutter in `Yoti.Auth.Images.Image`?

bucky-boy: Either _Image_, _BaseImage_, _ImageBase_ or _AbstractImage_ is fine.
Maybe not _AbstractImage_ - it's a bit Picasso.
.Net has a convention of using _xxxBase_, but there's no need for you to stack to that.

src/Yoti.Auth/Activity.cs
```diff
@@ -82,11 +83,10 @@ internal void SetAddressToBeFormattedAddressIfNull()
 
                 if (formattedAddressJToken != null)
                 {
-                    var addressValue = new YotiAttributeValue(YotiAttributeValue.TypeEnum.Text, formattedAddressJToken.ToString());
-
                     var addressAttribute = new YotiAttribute<string>(
                         name: Constants.UserProfile.PostalAddressAttribute,
-                        value: addressValue,
+                        type: AttrpubapiV1.ContentType.String,
```
bucky-boy: Surely we don't need this protobuf _Type_ in here....?

src/Yoti.Auth/Images/Image.cs
echarrod: I prefer ImageBase anyway, so will change to that!

src/Yoti.Auth/Aml/AmlException.cs
```diff
@@ -4,6 +4,10 @@ namespace Yoti.Auth.Aml
 {
     public class AmlException : Exception
     {
+        public AmlException()
```
bucky-boy: Do we need this?  An _Exception_ with absolutely nothing in it is not terribly informative

src/Yoti.Auth/Images/Image.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        private readonly AttrpubapiV1.ContentType _type;
```
bucky-boy: Concrete subclasses will know the Type of Image they represent, so this _Type_ property should not be required.

And we'd be better off removing references to any protobuf generated types.  Try to keep them contained in the converters - don't let the dependency spread around if possible, even if we don't expose it.

src/Yoti.Auth/Images/Image.cs
echarrod: We do have `BaseAttribute` currently though, so I'm thinking `BaseImage` would be more consistent, what do you think?

echarrod: Have changed to `Image`, as it's not necessary to convey it's a base class to clients.

src/Yoti.Auth/Aml/AmlException.cs
echarrod: VS suggested to implement this argument-less constructor if it was implementing `Exception`, have changed it to implement the base class in this scenario.

bucky-boy: Ok, but I don't see why it wants to define a constructor that we're not going to use.  Especially as it's a fairly unhelpful one, that may encourage us to throw an unhelpful exception.

See here: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions

They want us to define 4x constructors, and also make it _serialisable_.  I think they're wrong - we shouldn't add stuff we don't intend to use.

src/Yoti.Auth/AttributeConverter.cs
```diff
@@ -70,5 +64,26 @@ private static List<Yoti.Auth.Anchors.Anchor> ParseAnchors(AttrpubapiV1.Attribut
...
+            Dictionary<string, JToken> deserializedJson = Newtonsoft.Json.JsonConvert.DeserializeObject<Dictionary<string, JToken>>(utf8JSON);
+            return deserializedJson;
+        }
+
+        private static DateTime GetDateValue(byte[] bytes)
+        {
+            if (DateTime.TryParseExact(
+                        s: Conversion.BytesToUtf8(bytes),
+                        format: "yyyy-MM-dd",
+                        provider: CultureInfo.InvariantCulture,
```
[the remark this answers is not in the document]
echarrod: Have added comment to SDK-694, we can pick it up there

src/Yoti.Auth/AttributeConverter.cs
```diff
@@ -1,54 +1,48 @@
...
-                        attribute.Name,
-                        value,
-                        ParseAnchors(attribute));
+                    return new YotiAttribute<Dictionary<string, JToken>>(
+                      attribute.Name,
+                      value: GetJsonValue(attribute.Value.ToByteArray()),
+                      ParseAnchors(attribute));
 
                 case ContentType.Undefined:
                     // do not return attributes with undefined content types
```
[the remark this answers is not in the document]
echarrod: Will do that in the next PR 👍

src/Yoti.Auth/BaseAttribute.cs
```diff
@@ -13,29 +13,29 @@ namespace Yoti.Auth
...
     public abstract class BaseAttribute
     {
-        private readonly string _name;
-        private List<Anchor> _anchors;
-
-        private protected BaseAttribute(string name, YotiAttributeValue value)
+        private readonly string _name;
+        private List<Anchor> _anchors;
+
+        private protected BaseAttribute(string name)
```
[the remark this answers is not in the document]
echarrod: Nope, good spot

README.md
```diff
@@ -118,7 +118,7 @@ if (activityDetails.Outcome == ActivityOutcome.Success)
     if (user != null)
     {
         Image selfie = profile.Selfie.GetImage();
-        string selfieURI = profile.Selfie.GetBase64URI();
+        string selfieURI = profile.Selfie.Base64URI;
```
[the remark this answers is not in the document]
echarrod: Ah yes, changed it without updating this

src/Yoti.Auth/AttributeConverter.cs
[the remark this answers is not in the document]
echarrod: Have added SDK-697 for this

PR #63, opened 2018-12-05
SDK-697 SDK-667 SDK-692: Undefined content types, examples, SDK Version

echarrod:
1) Move Example projects to Yoti.Auth solution, tidy up
1) Send SDK version in request header -I've done this with reflection, so that there's not another place that we update the version, in case we would ever forget. If for any reason the version can't be retrieved, it just sets it to "Unknown", although I can't think of any scenarios where this might happen
1) Change attribute processing to add new content types as YotiAttribute<object>. These can then be retrieved in the normal way, but the value will need to be cast to the desired type
1) Added tests around GetAttribute (by name, without specifying the attribute type). This also returns a YotiAttribute<object>, which again can be cast


src/Yoti.Auth/AttributeConverter.cs
```diff
@@ -49,7 +49,10 @@ public static BaseAttribute ConvertToBaseAttribute(AttrpubapiV1.Attribute attrib
                     return null;
 
                 default:
-                    return null;
+                    return new YotiAttribute<object>(
```
bucky-boy: Should be YotiAttribute\<string\>.
The underlying value type is a string, and will always be a string.  It is a string representation of some other type we don't recognise, I grant you, but it is a string nonetheless.

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -111,11 +112,18 @@ public async Task<AmlResult> PerformAmlCheckAsync(string appId, AsymmetricCipher
             if (string.IsNullOrEmpty(authDigest))
                 throw new InvalidOperationException("Could not sign request");
 
+            string SDKversion = typeof(YotiClientEngine).GetTypeInfo()?.Assembly?.GetName()?.Version?.ToString();
+            if (SDKversion == null)
```
bucky-boy: I'm not clear why this would happen.
Surely it either gets the version from the assembly successfully, or it doesn't.....?

echarrod: The `?` is the "Safe Navigation Operator", which will return null if at any stage it is unable to find the requested object, rather than throwing an exception. 

Even though I don't know why this would happen either. I thought it would be preferable to add this, so that we wouldn't be throwing an exception and failing the request in the event it fails, as I'm sure we wouldn't want to stop the request from going through if there's any potential problems retrieving this version.

src/Yoti.Auth/AttributeConverter.cs
```diff
@@ -49,7 +49,10 @@ public static BaseAttribute ConvertToBaseAttribute(AttrpubapiV1.Attribute attrib
                     return null;
 
                 default:
-                    return null;
```
sidibos: Would there be any case when content type is none of the above.
I thought if photo doesn't know the content type, it will be set to Undefined. 
I am not sure we need this case here.

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -111,11 +112,18 @@ public async Task<AmlResult> PerformAmlCheckAsync(string appId, AsymmetricCipher
             if (string.IsNullOrEmpty(authDigest))
                 throw new InvalidOperationException("Could not sign request");
 
+            string SDKversion = typeof(YotiClientEngine).GetTypeInfo()?.Assembly?.GetName()?.Version?.ToString();
```
sidibos: The variable name should be in camelCase `SDKVersion`?

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -111,11 +112,18 @@ public async Task<AmlResult> PerformAmlCheckAsync(string appId, AsymmetricCipher
...
+            if (SDKversion == null)
+            {
+                SDKversion = "Unknown";
```
sidibos: Will it be useful to send version value as `Unknown`.
If there is an error, I would rather log it and not send the value as `Unknown`, your call.

src/Yoti.Auth/AttributeConverter.cs
echarrod: If a new content type is added, e.g. `Bitmap = 6`, then if we haven't updated the SDKs then I believe it would go down this route. 

If I remember correctly, `Undefined` content types are only used if there's been an error. @bucky-boy do you remember? 

In any case, we might not need it (or we might be very unlikely to need it), but is there any harm in leaving it here?

src/Yoti.Auth/YotiClientEngine.cs
echarrod: Good spot

src/Yoti.Auth/YotiClientEngine.cs
echarrod: If we're logging an error, it's not likely we'll be able to see it. I guess it wouldn't harm to also log it though, I'll add it. 

If we send the header as "Unknown", then we will be able to see it in the logs.

src/Yoti.Auth/AttributeConverter.cs
sidibos: 👍

src/Yoti.Auth/YotiClientEngine.cs
sidibos: I guess you could see it when you run your test cases locally and will be helpful.
It's your call.

src/Yoti.Auth/YotiClientEngine.cs
bucky-boy: We'll be better off determining whether it happens or not, and if so why.

If there is no reason for it to be there, it should not be there.

PR #64, opened 2018-12-14
[SDK-575]: Age Verifications

echarrod:
- Add Age Verifications
- Add [Microsoft.Fakes](https://docs.microsoft.com/en-us/visualstudio/test/isolating-code-under-test-with-microsoft-fakes?view=vs-2017) as mocking framework, which allows us to stub/shim method, this required changing the framework of the tests from .NET Core, as MS Fakes is not supported in that yet.


src/Yoti.Auth/YotiProfile.cs
```diff
@@ -147,5 +153,36 @@ public YotiAttribute<DocumentDetails> DocumentDetails
...
+            {
+                return _ageVerificationParser.FindAllAgeVerifications();
+            }
+        }
+
+        /// <summary>
+        /// Searches for an <see cref="AgeVerification"/> corresponding to an `Age Under` check for the given age.
+        /// Returns null if no match was found.
+        /// </summary>
+        public AgeVerification AgeUnderVerification(int age)
```
bucky-boy: Again, maybe this should be _FindAgeUnderVerification_.

Maybe my brain is too Java-wired now, but this just doesn't look right.

src/Yoti.Auth/Verifications/AgeVerificationParser.cs
```diff
@@ -0,0 +1,75 @@
...
+                FindAgeOverVerifications();
+
+                _allVerificationsDict = _ageUnderVerificationsDict.Values
+                    .Concat(_ageOverVerificationsDict.Values).ToList();
+            }
+
+            return _allVerificationsDict.AsReadOnly();
+        }
+
+        public AgeVerification AgeUnderVerification(int age)
```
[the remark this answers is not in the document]
echarrod: Will rename this to `FindAgeUnderVerification`, and change `FindAgeOverVerifications` to `FindAllAgeOverVerifications`, sound ok?

src/Yoti.Auth/YotiProfile.cs
echarrod: Same as above

src/Yoti.Auth/Verifications/AgeVerificationParser.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public AgeVerification FindAgeOverVerification(int age)
+        {
+            return _ageOverVerificationsDict?.FirstOrDefault(
+                x => x.Key == string.Format("{0}:{1}", Constants.UserProfile.AgeOverAttribute, age))
+                .Value;
+        }
+
+        private static Dictionary<string, AgeVerification> FindAgeUnderVerifications(BaseProfile baseProfile)
```
bucky-boy: We can refactor this to work for both _age_under_ and _age_over_, if we pass in a value for _Constants.UserProfile.AgeOverAttribute_ (or _AgeUnderAttribte_).  Then we won't need two methods for the same task.

src/Yoti.Auth/Verifications/AgeVerificationParser.cs
```diff
@@ -0,0 +1,65 @@
...
+            _ageUnderVerificationsDict = FindAgeUnderVerifications(baseProfile);
+            _ageOverVerificationsDict = FindAgeOverVerifications(baseProfile);
+
+            _allVerificationsDict = _ageUnderVerificationsDict.Values
+                .Concat(_ageOverVerificationsDict.Values).ToList();
+        }
+
+        public ReadOnlyCollection<AgeVerification> FindAllAgeVerifications()
+        {
+            return _allVerificationsDict.AsReadOnly();
```
bucky-boy: We can make this _AsReadOnly_ call in the constructor - so then we don't have to create a new instance each time it's called

src/Yoti.Auth/Verifications/AgeVerificationParser.cs
echarrod: SOLI**D** suggestion

PR #66, opened 2018-12-19
Code Analysis

echarrod:
- Add FXCop Analyzer
- Mark protobuf files as auto-generated, so they won't be picked up by Code Analysis (pragma disables are then no longer necessary)


src/Yoti.Auth/Anchors/Anchor.cs
```diff
@@ -40,7 +40,7 @@ public Anchor(AttrpubapiV1.Anchor protobufAnchor)
             _signedTimeStamp = new SignedTimestamp(protobufSignedTimestamp);
         }
 
-        private List<X509Certificate2> ConvertRawCertToX509List(RepeatedField<ByteString> rawOriginServerCerts)
+        private static List<X509Certificate2> ConvertRawCertToX509List(RepeatedField<ByteString> rawOriginServerCerts)
```
bucky-boy: Not sure how this will turn out in C#, but in Java static code causes problems for unit testing - it's harder to mock it

src/Examples/CoreExample/Controllers/AccountController.cs
```diff
@@ -15,7 +15,17 @@ public AccountController(ILogger<AccountController> logger)
             _logger = logger;
         }
 
-        public static byte[] PhotoBytes { get; set; }
+        private static byte[] photoBytes;
```
bucky-boy: Not convinced this should be static

src/Yoti.Auth/Anchors/Anchor.cs
echarrod: We can overcome this using shims & MS.Fakes 👍

echarrod:
The analyzer is run on build, so nothing needs to be changed. I will add it to the .NET wiki though, and how/when to suppress the issues.

src/Examples/CoreExample/Controllers/AccountController.cs
echarrod: You're right, will change.

PR #67, opened 2018-12-20
[SDK-720]: Remove ActivityDetails.Outcome

echarrod:
- In favour of throwing exceptions instead


src/Yoti.Auth/YotiProfileException.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace Yoti.Auth
+{
+    internal class YotiProfileException : Exception
+    {
+        public YotiProfileException()
```
bucky-boy: I know MS suggest that you should always implement all three constructors, but maybe this in an occassion where we should ignore them.

Creating an Exception using its default, no-arg constructor is probably the most unhelpful thing we could ever do.  We shouldn't do it.  And since we have no intention of using it, it shouldn't exist.  Any time we want to throw an exception, we should require a _message_ as a minimum.

Googling it, the only justification I've found for MS's position is the _principal of least surprise_ - if Exceptions are always defined with these three constructors, other developers will always find them easy to use.
1. This Exception is for us SDK developers to throw, not our SDK consumers
2. If you stop and think it through, MS's position is incredibly daft.  They want all devs everywhere to be able to throw non-descriptive exceptions for the sake of easiness (a bad practice in itself), and they're encouraging the thoughtless, blind application of boiler-plate code throughout the application

What happens if we take it out?  Can't we just suppress any warnings we get for not following MS policy?

echarrod: I guess it _might_ be useful if the name of the exception is descriptive enough. That's not the case here, so I'll remove it and add a suppression.

PR #72, opened 2019-04-11
SDK-914: Address FXCop Warnings

echarrod:
I split each commit into looking at separate issues (and sometimes added other small issues into each commit too). As such, it would probably be best to review this by looking at the individual commits. 

- [CA2007](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2007-do-not-directly-await-task?view=vs-2019) ConfigureAwait - defaults to configureAwait(true). I have changed this to false in the instances where we want it to be async, and set it to true in the methods where we are calling an async method from a synchronous method. 
- [CA1305](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1305-specify-iformatprovider?view=vs-2019) - specifying behaviour so the code won't behave differently in different locales
- S3925 & CA2237: these methods FXCop complains about are available in all of the frameworks except `netstandard` (issue on Github - https://github.com/dotnet/coreclr/issues/2715#event-519435422 ), as it seems Microsoft extracted it to a [NuGet package](https://www.nuget.org/packages/System.Runtime.Serialization.Formatters/). This [SO](https://stackoverflow.com/questions/39199049/what-is-the-equivalent-of-serializable-in-net-core-conversion-projects) summarises it quite nicely. We can satisfy this complaint by adding the NuGet package and implementing the extra constructors, but I'm not a fan of adding a NuGet package for something that we're never going to use, and doesn't seem necessary. 
- The rest I think are more straightforward, but feel free to comment on any of them that don't make sense.


src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -70,17 +70,24 @@ public async Task<AmlResult> PerformAmlCheckAsync(string appId, AsymmetricCipher
                 throw new ArgumentNullException(nameof(amlProfile));
             }
 
+            return PerformAmlCheckInternalAsync(appId, keyPair, apiUrl, amlProfile);
+        }
+
+        public async Task<AmlResult> PerformAmlCheckInternalAsync(string appId, AsymmetricCipherKeyPair keyPair, string apiUrl, IAmlProfile amlProfile)
```
davidgrayston: Should this be private?

echarrod: Good spot

PR #74, opened 2019-04-17
SDK-721: Exceptions

echarrod:
Dynamic Scenarios is getting quite large so I'm trying to check in some other parts before the main chunk to lessen the load:
- Make CreateExceptionFromStatusCode generic for reuse
- Move exceptions to own folder (looks like `git mv` doesn't work when you also make changes to the file 😒 )


src/Yoti.Auth/GlobalSuppressions.cs
```diff
@@ -4,10 +4,10 @@
 // a specific target and scoped to a namespace, type, member, etc.
 
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2237:Mark ISerializable types with serializable", Justification = "Serializable not available with NET Standard framework", Scope = "type", Target = "~T:Yoti.Auth.Aml.AmlException")]
```
davidgrayston:
```suggestion
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2237:Mark ISerializable types with serializable", Justification = "Serializable not available with NET Standard framework", Scope = "type", Target = "~T:Yoti.Auth.Exceptions.AmlException")]
```

PR #78, opened 2019-05-03
SDK-893: Update UNDEFINED Content type handling

echarrod:
- Testing that the remaining attributes continue to be parsed required a small amount of reorganising.
- I also noticed there was no test coverage for the exceptions in ProfileParser, so I added those. 
- Using [NLog](https://nlog-project.org/) in favour of other logging tools as it is open-source, seems to be the de facto standard, and is actively maintained. It also seems to be the [most efficient](https://www.loggly.com/blog/benchmarking-5-popular-net-logging-libraries/). We are using the MS logging in the example projects, so I will look at swapping that out in future.

### DIFF
--- a/test/Yoti.Auth.Tests/AttributeConverterTests.cs
+++ b/test/Yoti.Auth.Tests/AttributeConverterTests.cs
@@ -1,0 +1,41 @@
+﻿using Google.Protobuf;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Yoti.Auth.Tests
+{
+    [TestClass]
+    public class AttributeConverterTests
+    {
+        private const string StringValue = "{}";
+        private readonly ByteString _byteValue = ByteString.CopyFromUtf8(StringValue);
+
+        [TestMethod]
+        public void FailureInAttributeParsing_ShouldNotStopOtherAttributes()
+        {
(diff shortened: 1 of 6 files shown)


test/Yoti.Auth.Tests/AttributeConverterTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        private const string StringValue = "{}";
+        private readonly ByteString _byteValue = ByteString.CopyFromUtf8(StringValue);
+
+        [TestMethod]
+        public void FailureInAttributeParsing_ShouldNotStopOtherAttributes()
+        {
+            var attribute1 = CreateProtobufJsonAttribute(Constants.UserProfile.StructuredPostalAddressAttribute, _byteValue);
+            var invalidAttribute = CreateProtobufJsonAttribute(Constants.UserProfile.PostalAddressAttribute, ByteString.CopyFromUtf8("invalid"));
+            var attribute2 = CreateProtobufJsonAttribute(Constants.UserProfile.PhoneNumberAttribute, _byteValue);
```
davidgrayston: Could we use descriptive attribute names here, something like `valid 1`, `valid 2` and `invalid`. It's fine as is, but might make the test easier to follow?

echarrod: Good point. I'm actually editing this file in another PR, so will change it in that 👍

PR #79, opened 2019-05-03
SDK-857: Handle empty attribute values

echarrod:


src/Yoti.Auth/AttributeConverter.cs
```diff
@@ -36,6 +36,10 @@ public static BaseAttribute ConvertToBaseAttribute(ProtoBuf.Attribute.Attribute
         {
             NLog.Logger logger = NLog.LogManager.GetCurrentClassLogger();
 
+            if (attribute.ContentType != ContentType.String
+                    && attribute.Value.Length == 0)
+                throw new InvalidOperationException("Empty value is invalid for non-string content types");
```
davidgrayston: Could we add test coverage for this exception? Or perhaps this method should be private as it's only being used by `Yoti.Auth.Tests.ActivityTests::AddAttributeToProfile()`?

echarrod: In `OtherAttributesWithEmptyValueAreNotAdded` we're testing that each of the attribute types apart from string are causing the attribute to not be added when the empty value it attempted to be added. I could add further tests for this method though, to check that this exact exception is being thrown.

echarrod: Also the class is `internal`, so this method would only inherit the `internal` access modifier.

echarrod: @davidgrayston have added some further tests in a098202

PR #80, opened 2019-05-08
Sdk 477 docker demo

echarrod:
I've added this to the public repo so that clients have a way of easily running the demo project too. 

I first went down the dockerfile route, without docker-compose. Quite a lot of config arguments are required to make it work though, and the VS-generated docker-compose file seems to be the only one which also works OOTB when run on the command line (dockerfile doesn't). 

I would have liked to have put the docker-compose file in the Examples folder, but we need to also pass the local Yoti package in with the build context, if we are to get that version, and not the one from NuGet. 

To get the Docker Support to work, I had to recreated the project in VS, which added the appropriate references & configuration, and altered the setup slightly to match the latest MS recommendations. 

There also seemed to be some issues with port clashing, so I changed the port for when running in Docker to be 44380. 

Run the docker instructions in the README to let me know if that's everything that you need to get it going 👍


README.md
```diff
@@ -212,16 +212,24 @@ bool onWatchList = amlResult.IsOnWatchList();
...
 1) Fill in the environment variables in this file with the ones specific to your application (mentioned in the [Client initialisation](#client-initialisation) section)
 1) Restore the static files needed with `bower install`
 1) From the Yoti Dashboard, set the application domain to `localhost:44344` and the scenario callback URL to `/account/connect`
+
+##### With Docker
+1) Make sure the .pem file is within the `src\Examples\CoreExample` folder.
+1) Rename it to be `YotiKey.pem`
+1) `cd src`
+1) `docker-compose build --no-cache`
+1) `docker-compose up`
```
davidgrayston: I received the following message when running `docker-compose up`:
```
WARNING: The APPDATA variable is not set. Defaulting to a blank string.
```

Tried removing the following lines fro `docker-compose.override.yml`:
```
  - ${APPDATA}/ASP.NET/Https:/root/.aspnet/https:ro
  - ${APPDATA}/Microsoft/UserSecrets:/root/.microsoft/usersecrets:ro
```

But then outputs message:
```
coreexample_1  | crit: Microsoft.AspNetCore.Server.Kestrel[0]
coreexample_1  |       Unable to start Kestrel.
```

Is there a way we can include these without using a volume on the host machine?

echarrod:
@davidgrayston I changed the container to be Windows-based, which gets round the problem of having to mount the secrets file and certificate. It does seem like it's not easy to set up windows containers on [Mac](https://forums.docker.com/t/how-do-i-start-a-windows-docker-container-on-my-mac-os-x/12953) or [Linux](https://stackoverflow.com/questions/42158596/can-windows-containers-be-hosted-on-linux). Since this was the reason I was including this in the PR, I'm thinking of moving this to run only on Concourse, and it won't be in the public repo.

echarrod:
It doesn't seem easy or sensible to set up Windows containers to work with Concourse. I've reverted to linux containers, with the certificate being checked in. Let me know if this now works for you @davidgrayston

echarrod:
We need to send the `src` folder in the build context with Docker. We could mount the volume, but I think it's fine to leave as is. When we publish the package, it is just the contents of the Yoti.Auth project file, so we're not in any danger of anything else being included

echarrod:
@davidgrayston I've moved the docker-compose to be within the CoreExample folder in 5c9c435

PR #81, opened 2019-05-09
SDK-784 Packages

echarrod:
- Package updating managed through NuGet
- Some ambiguous references, so had to add some explicit references too (`Microsoft.CodeAnalysis....`)
- Also removed unused references from 45Example project
- There were different versions of System.Net.Http being referenced, so i removed the older one, which was also using conditional framework specifications, which weren't neccesary


src/Yoti.Auth/CryptoEngine.cs
```diff
@@ -66,7 +65,7 @@ public static byte[] SignDigest(byte[] digestBytes, AsymmetricCipherKeyPair keyp
 
         public static byte[] GetDerEncodedPublicKey(AsymmetricCipherKeyPair keypair)
         {
-            return SubjectPublicKeyInfoFactory.CreateSubjectPublicKeyInfo(keypair.Public).GetDerEncoded();
+            return Org.BouncyCastle.X509.SubjectPublicKeyInfoFactory.CreateSubjectPublicKeyInfo(keypair.Public).GetDerEncoded();
```
davidgrayston: Could this also be added as `using Org.BouncyCastle.X509.SubjectPublicKeyInfoFactory;`?

Looks fine as is though

echarrod: You can't I'm afraid - `SubjectPublicKeyInfoFactory` is a type, and you can only import namespaces with `using`s. I would be tempted to reinstate `using Org.BouncyCastle.X509` if we were using it in more than one place, but I will leave it as it is since it's only one.

PR #82, opened 2019-05-15
SDK-497: Update Example projects with new design

echarrod:
Please try running the examples to check all of the necessary files are checked in to source control.

I had a bit of a nightmare trying to get both projects to reference a shared project and use the static files, views and controllers from that. That was in the [SDK-497](https://github.com/getyoti/yoti-dotnet-sdk/tree/SDK-497-Examples) branch. 
The projects seem to get a little confused when the files they are referencing aren't within the same directory. I needed to mark the files as "links" to the shared projects' files. BUT, when you do this, the files aren't copied to the built output, so I had to add the manual step in the .csproj to do this. This would have been ok, if a little hacky, but it then didn't work when we tried on Linux.

This PR is instead using two copies of the files, one for the .NET 4.5 project, and one for the .NET Core project. I will try again at having them both use a shared set of static files in the future - SDK-963

In order to have the example projects look through the attributes and decide how to display them, I had to make the constants and the attribute list public. I think we were thinking of doing this anyway. I don't see any other way for clients to easily look through the attributes like this.


src/Examples/45Example/Views/Account/Connect.cshtml
```diff
@@ -1,447 +1,95 @@
...
-                                string AnchorType = anchor.GetAnchorType().ToString();
-                                @Html.DisplayFor(m => AnchorType);
-                            }
-                        </td>
-                        <td rowspan="@Model.StructuredPostalAddress.GetValue().Count">
-                            @{
-                                string AnchorValue = string.Join(" , ", anchor.GetValue());
-                                @Html.DisplayFor(m => AnchorValue);
+        <section class="yoti-attributes-section">
+            <img class="yoti-company-logo" src="~/static/assets/company-logo.jpg" alt="company logo" />
```
davidgrayston: Should this show the real company logo if available? The PHP example is showing the static image too so perhaps this can be fixed in a separate task

echarrod: @KiranBali I don't suppose you know why we have the template logo and not the actual application profile's logo in the sdk-static-pages designs? Do you foresee any problems if we change it to use the logo if present, and if not then use this template logo? 
image

echarrod: I spoke to Kiran, and she doesn't see too much value in changing this - as it's just for a demo. We'll leave it as is

PR #84, opened 2019-05-16
SDK-694: DateTimes & Descriptions

echarrod:
- See ticket SDK-694 for more info - have opted to not make any other changes as part of that ticket
- Added to descriptions of DateTime objects to explain the times will default to 00:00:00
- Added remaining DocumentDetails descriptions - please review

### DIFF
--- a/src/Yoti.Auth/Document/DocumentDetails.cs
+++ b/src/Yoti.Auth/Document/DocumentDetails.cs
@@ -6,7 +6,7 @@ namespace Yoti.Auth.Document
     public class DocumentDetails
     {
         /// <summary>
-        /// 3 digit country code, e.g. “GBR“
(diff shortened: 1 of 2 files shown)


src/Yoti.Auth/Document/DocumentDetails.cs
```diff
@@ -5,10 +5,31 @@ namespace Yoti.Auth.Document
 {
     public class DocumentDetails
     {
+        /// <summary>
+        /// 3 digit country code, e.g. “GBR“
```
davidgrayston: Should this be "alpha-3" or "3 letter"?

echarrod: Good catch, changed in cba58b0

PR #89, opened 2019-05-31
SDK-980 Dynamic Scenarios - optional attributes

echarrod:
Since it wasn't possible to make everything private which used the optional functionality, I've removed the optional parts for now. We can revert this change when optional attributes are enabled in the app.

### DIFF
--- a/src/Yoti.Auth/ShareUrl/Policy/DynamicPolicy.cs
+++ b/src/Yoti.Auth/ShareUrl/Policy/DynamicPolicy.cs
@@ -17,19 +17,14 @@ namespace Yoti.Auth.ShareUrl.Policy
         [JsonProperty(PropertyName = "wanted_remember_me")]
         private readonly bool _wantedRememberMeId;
 
-        [JsonProperty(PropertyName = "wanted_remember_me_optional")]
-        private readonly bool _isWantedRememberMeIdOptional;
-
         public DynamicPolicy(
             ICollection<WantedAttribute> wantedAttributes,
             List<int> wantedAuthTypes,
-            bool wantedRememberMeId,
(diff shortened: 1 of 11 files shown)


src/Yoti.Auth/ShareUrl/Policy/DynamicPolicy.cs
```diff
@@ -17,19 +17,14 @@ public class DynamicPolicy
         [JsonProperty(PropertyName = "wanted_remember_me")]
         private readonly bool _wantedRememberMeId;
 
-        [JsonProperty(PropertyName = "wanted_remember_me_optional")]
```
davidgrayston: As you already mentioned to me, we may want to keep this property and set to `false`

test/Yoti.Auth.Tests/TestData/DynamicPolicy.json
```diff
@@ -5,27 +5,26 @@
...
       {
         "name": "date_of_birth",
         "derivation": "age_under:40",
-        "optional": true
+        "optional": false
       }
     ],
     "wanted_auth_types": [ 2 ],
-    "wanted_remember_me": false,
-    "wanted_remember_me_optional": false
```
davidgrayston: We'd need to reinstate the `wanted_remember_me_optional` property here if you set to `false`

echarrod: Yep, updated in 179b5be

src/Yoti.Auth/ShareUrl/Policy/DynamicPolicy.cs
echarrod: I've now done that, completed in 179b5be

PR #94, opened 2019-06-25
SDK-468: Wrapper App changes

echarrod:
- Add DocumentDetails to README
- Make Yoti API configurable, for use with the wrapper application
- Creating URIs seem to sometimes result in a trailing slash, which caused a double slash when combined with the endpoint. I created `CreateRequestUri` to remove this trailing slash if it's present
- Made ApplicationProfile consts public (like UserProfile), so we can reference them externally

### DIFF
--- a/src/Yoti.Auth/YotiClientEngine.cs
+++ b/src/Yoti.Auth/YotiClientEngine.cs
@@ -43,15 +43,29 @@ namespace Yoti.Auth
 
             Dictionary<string, string> headers = HeadersFactory.Create(keyPair, httpMethod, endpoint, httpContent: null);
 
+            Uri requestUri = CreateRequestUri(apiUrl, endpoint);
+
             Response response = await _httpRequester.DoRequest(
                 _httpClient,
                 HttpMethod.Get,
-                new Uri(
-                    apiUrl + endpoint),
+                requestUri,
                 headers,
                 httpContent).ConfigureAwait(false);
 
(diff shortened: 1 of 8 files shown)


src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -43,17 +43,31 @@ public async Task<ActivityDetails> GetActivityDetailsAsync(string encryptedConne
...
 
             return ProfileParser.HandleResponse(keyPair, response);
         }
 
+        private static Uri CreateRequestUri(Uri apiUrl, string endpoint)
+        {
+            string baseApiUrl = apiUrl.ToString();
+            if (baseApiUrl.EndsWith("/", StringComparison.OrdinalIgnoreCase))
+            {
+                return new Uri(baseApiUrl.TrimEnd('/') + endpoint);
```
davidgrayston: Do we need to check for a trailing slash before calling `TrimEnd('/')`? It seems like we can always trim `baseApiUrl ` based on the docs:
>If no characters can be trimmed from the current instance, the method returns the current instance unchanged.

So the resulting code would be:
```c#
private static Uri CreateRequestUri(Uri apiUrl, string endpoint)
{
    return new Uri(apiUrl.ToString().TrimEnd('/')  + endpoint);
}
```

echarrod: Good spot, have changed in 5b31a24

PR #95, opened 2019-07-03
SDK-1046: Adding AML Example

echarrod:
- General tidying of README - will copy changes to dev docs
- Added comment to `Attributes` to clarify usage
- Removed redundant methods & update tests
- Moved profile examples into separate folder
- Added AML Example - this can be tried using .NET Core on Mac/Linux

Example output in console:
```bash
Yoti AML Example

Using environment variables from .env file:
sdkId='19e50ec8-5ef9-4974-a7a9-dac7ad2a700a'
yotiKeyFilePath='C:\Keys\NET-Core-access-security.pem'


Completing check for AML profile: '{
  "given_names": "Edward Richard George",
  "family_name": "Heath",
  "ssn": null,
  "address": {
    "post_code": null,
    "country": "GBR"
  }
}'


Result:
Is on PEP list: False
Is on Fraud list: False
Is on Watch list: False
```


src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -25,13 +25,6 @@ public YotiClientEngine(IHttpRequester httpRequester, HttpClient httpClient)
 #endif
         }
 
-        public ActivityDetails GetActivityDetails(string encryptedToken, string sdkId, AsymmetricCipherKeyPair keyPair, Uri apiUrl)
```
davidgrayston: Is it convention to suffix async methods with `...Async`? If not, could we make `GetActivityDetails` async and remove `GetActivityDetailsAsync`?

src/Yoti.Auth/YotiClientEngine.cs
```diff
@@ -53,13 +46,6 @@ public async Task<ActivityDetails> GetActivityDetailsAsync(string encryptedConne
             return ProfileParser.HandleResponse(keyPair, response);
         }
 
-        public AmlResult PerformAmlCheck(string appId, AsymmetricCipherKeyPair keyPair, string apiUrl, IAmlProfile amlProfile)
```
davidgrayston: Same, could `PerformAmlCheck` be async and remove `PerformAmlCheckAsync`?

src/Yoti.Auth/YotiClientEngine.cs
echarrod: Yes it's a .NET convention.

From <https://docs.microsoft.com/en-us/dotnet/csharp/async>: 

> You should add "Async" as the suffix of every async method name you write.
> This is the convention used in .NET to more-easily differentiate synchronous and asynchronous methods. Note that certain methods which aren’t explicitly called by your code (such as event handlers or web controller methods) don’t necessarily apply. Because these are not explicitly called by your code, being explicit about their naming isn’t as important.

There is some disagreement([1](https://stackoverflow.com/a/47810905/1014632) | [2](https://docs.particular.net/nservicebus/upgrades/5to6/async-suffix)) on this if your code doesn't contain both async and sync methods (which is the case here), and it's also not in a public class, but in general I think it would be best to go with the convention.

PR #96, opened 2019-07-05
SDK-1046: Add USA example

echarrod:
- Choice of whether to use USA address is configured with an env variable


echarrod:
As discussed, although it's more consistent if we prefix with `YOTI_`, I don't think it's as clear. Will leave as is

PR #98, opened 2019-07-08
SDK-992: Update attribute descriptions

echarrod:
- I used an autoformatter for XML comments too, to ensure they don't go over a certain width.

### DIFF
--- a/src/Yoti.Auth/Profile/YotiProfile.cs
+++ b/src/Yoti.Auth/Profile/YotiProfile.cs
@@ -41,7 +41,7 @@ namespace Yoti.Auth.Profile
 
         /// <summary>
         /// FullName represents the user's full name. If family_name/given_names are present, the
-        /// value will be equal to the string 'given_names + " " family_name'. Will be nil if not
+        /// value will be equal to the string 'given_names + " " + family_name'. Will be nil if not
         /// provided by Yoti.
         /// </summary>
         public YotiAttribute<string> FullName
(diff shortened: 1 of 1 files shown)


src/Yoti.Auth/Profile/YotiProfile.cs
```diff
@@ -40,7 +40,9 @@ public YotiAttribute<Image> Selfie
         }
 
         /// <summary>
-        /// FullName represents the user's full name. This will be null if not provided by Yoti.
+        /// FullName represents the user's full name. If family_name/given_names are present, the
+        /// value will be equal to the string 'given_names + " " family_name'. Will be nil if not
```
davidgrayston:
```suggestion
        /// value will be equal to the string 'given_names + " " + family_name'. Will be nil if not
```

Should there be an extra `+` here?

src/Yoti.Auth/Profile/YotiProfile.cs
```diff
@@ -95,7 +101,8 @@ public YotiAttribute<string> EmailAddress
         }
 
         /// <summary>
-        /// DateOfBirth represents the user's date of birth as a DateTime. The time part of this DateTime will default to 00:00:00. This will be null if not provided by Yoti.
+        /// DateOfBirth represents the user's date of birth. Will be nil if not provided by Yoti. The
+        /// time part of this DateTime will default to 00:00:00. Will be null if not provided by Yoti.
```
davidgrayston: There's a mixture of nil/null in the documentation - is that correct?

echarrod: Oh yes, corrected in bf8af66

PR #99, opened 2019-07-25
SDK-600: Sandbox

echarrod:
Actual Sandbox changes are getting quite large, so I'm going to create some preliminary PRs so that it's not one big one.

The XML comments are set to auto format now, so they have been updated to all be a fixed width, and if they are too long they will be chopped. This means it will probably be easier to view the changes by individual commits. It added a space before the s in `list of <see cref="anchor">s` (and other similar plurals), so I just removed the `s`'s.


src/Yoti.Auth/ShareUrl/Extensions/TransactionalFlowExtensionBuilder.cs
```diff
@@ -1,8 +1,9 @@
 ﻿namespace Yoti.Auth.ShareUrl.Extensions
 {
     /// <summary>
-    /// Allows you to provide a non-null object representing the content to be submitted in the TRANSACTIONAL_FLOW extension.
-    /// The object will be mapped to a Json representation using Jackson ObjectMapper.
+    /// Allows you to provide a non-null object representing the content to be submitted in the
+    /// TRANSACTIONAL_FLOW extension. The object will be mapped to a Json representation using
+    /// Jackson ObjectMapper.
```
davidgrayston: Is `Jackson ObjectMapper` reference from SDK Java?

echarrod: I understood this to mean that the code parsing the extension would be using Jackson ObjectMapper, so would be SDK-independent. @bucky-boy are you able to confirm this?

echarrod: Spoke with Vince, you were right! I've removed it in 5a2b971

bucky-boy: Yes, sorry I hadn't seen your question.  ObjectMapper is Java specific.
Besides, there's no point including this kind of detail in the docs anyway.  We should remove this from the java docs, it's not helpful

PR #102, opened 2019-07-31
SDK-1094: Allow removal of authType, ensure uniqueness

echarrod:
- WithPinAuthentication(enabled:false) / WithSelfieAuthentication(enabled:false) now removes that AuthType from the Policy, and WithAuthType now has `enabled` field, allowing other auth types to be removed too
- Ensuring uniqueness for AuthTypes by using HashSet instead of a List

### DIFF
--- a/src/Yoti.Auth/ShareUrl/Policy/DynamicPolicyBuilder.cs
+++ b/src/Yoti.Auth/ShareUrl/Policy/DynamicPolicyBuilder.cs
@@ -4,11 +4,8 @@ namespace Yoti.Auth.ShareUrl.Policy
 {
     public class DynamicPolicyBuilder
     {
-        private const int _selfieAuthType = 1;
-        private const int _pinAuthType = 2;
-
         private readonly Dictionary<string, WantedAttribute> _wantedAttributes = new Dictionary<string, WantedAttribute>();
-        private readonly List<int> _wantedAuthTypes = new List<int>();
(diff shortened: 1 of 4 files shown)


src/Yoti.Auth/ShareUrl/Policy/DynamicPolicyBuilder.cs
```diff
@@ -111,17 +108,21 @@ public DynamicPolicyBuilder WithSelfieAuthentication(bool enabled)
         {
             if (enabled)
             {
-                return WithAuthType(_selfieAuthType);
+                return WithAuthType(DynamicPolicy.SelfieAuthType);
             }
+
+            _wantedAuthTypes.Remove(DynamicPolicy.SelfieAuthType);
```
davidgrayston: Should this be in `WithAuthType()` with new enabled argument, i.e. `WithAuthType(int authType, bool enabled)`, or are we only allowing selfie and pin to be removed?

echarrod: Good point, I originally had `WithoutAuthType`, but thought that sounded silly! I think `WithAuthType(int authType, bool enabled)` seems sensible, and fitting with the rest of how we build dynamic policies.

echarrod: Added in 4be0e64

PR #103, opened 2019-08-01
SDK-1036: Add Modal to Example projects

echarrod:
- Removed APPLICATION_ID references also


src/Examples/Profile/45Example/Controllers/HomeController.cs
```diff
@@ -11,20 +11,26 @@ namespace Example.Controllers
 {
     public class HomeController : Controller
     {
-        private readonly string _appId;
+        private readonly string _yotiClientSdkId;
 
-        public ActionResult Index()
+        public HomeController()
```
echarrod: I just rearranged this to have the constructor at the top of the class

PR #104, opened 2019-08-01
SDK-1058: Add Structured Postal Address to example project

echarrod:
- Will also add document_images once it's been added to the SDK
- I added a generic `GetValue` method which returns an object, in case we decide to do custom displays for other attributes


echarrod:
@davidgrayston ah, I discarded the profile.css changes as I thought it was just spacing. I've updated in 84434ae

PR #105, opened 2019-08-06
SDK-992: Update sources & verifiers text

echarrod:


echarrod:
I asked Quirino about that this morning, he said it is a possible value if you decide to accept an `over_18` value when estimating an age with age scan.

PR #112, opened 2019-10-02
SDK-600: Sandbox Support

echarrod:
- Resources file has been added since FXCop complains if error messages are in English (this way they can be localized easily)
- Renamed `privateStreamKey` to `privateKeyStream`, unless anyone thinks `privateStreamKey` makes sense? 
- Allowed `YotiClient` to be initialised with an `AsymmetricCipherKeyPair`, which we make use of in the Sandbox. This is what the StreamReader (how the YotiClient is currently initialised) is converted into under the hood anyway, so just allows a different entrypoint.
- Removed the `apiUrl` specification from API requests (not yet released), in favour of using `OverrideApiUri` (so you only have to do it once) per Client.
- It's also recommended to specify a _minimum_ AspNetCore reference, so that's what the last file change is (`<PackageReference Include="Microsoft.AspNetCore.All" Version="2.1" />`)


test/Yoti.Auth.Sandbox.Tests/Profile/Request/Attribute/Derivation/SanxboxAgeVerificationTest.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System;
```
davidgrayston: This file is name "Sanxbox..."

echarrod: Good spot! Fixed in 2367ac4

PR #116, opened 2019-10-08
SDK-1225: Target frameworks

echarrod:
- Will wait for SDK v3.0.0 release before merging
- .NET Core 3.0 requires MS Build Tools 16 to compile, which is only included in Visual Studio 2019. I've chosen to not include this for now, as it will restrict who can compile the SDK (both internally and externally)
- Additionally Travis doesn't support .NET Core App 2.2 yet unfortunately, so I've only added 2.1

### DIFF
--- a/src/Yoti.Auth/Conversion.cs
+++ b/src/Yoti.Auth/Conversion.cs
@@ -31,8 +31,11 @@ namespace Yoti.Auth
         /// </summary>
         public static byte[] UrlSafeBase64ToBytes(string urlSafeBase64)
         {
+#if NETCOREAPP2_2
+            string base64 = urlSafeBase64.Replace("-", "+", StringComparison.Ordinal).Replace("_", "/", StringComparison.Ordinal);
+#else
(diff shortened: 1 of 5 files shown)


src/Yoti.Auth/Conversion.cs
```diff
@@ -31,8 +31,11 @@ public static byte[] Base64ToBytes(string base64)
         /// </summary>
         public static byte[] UrlSafeBase64ToBytes(string urlSafeBase64)
         {
+#if NETCOREAPP2_2
```
davidgrayston: Looks like 2.0 supports this https://docs.microsoft.com/en-us/dotnet/api/system.string.replace?view=netcore-2.0#System_String_Replace_System_String_System_String_System_StringComparison_ - should this be changed to `>= 2.0` or `2.1` to match _TargetFrameworks_ config

echarrod: Have changed this to have 2.0 & 2.1 in c140a77

PR #148, opened 2020-01-20
SDK-1381: Packages location

echarrod:
This seems to be a problem in that the "packages" folder is created differently on the Azure environment than on other machines. I think it might be a problem when using Visual Studio 2019, but I'm not sure of the root cause, it could be NuGet too. 

This change tries to import the package from both of the locations it could be present at, and I'm hoping this will be addressed and fixed so we can have consistent build environments both on Azure and elsewhere.

I reverted adding the static modifier, as this was causing a 403 error, I presume because it required to be non-static in order to be picked up.


src/Examples/Profile/47Example/47Example.csproj
```diff
@@ -23,6 +24,8 @@
     <IISExpressWindowsAuthentication />
     <IISExpressUseClassicPipelineMode />
     <UseGlobalApplicationHostFile />
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
```
davidgrayston: Could this be self closing? `<NuGetPackageImportStamp />`

echarrod: I'm sure it could be, but it was added by Visual Studio automatically, so I'm inclined to leave it as it is, so it's less likely to be changed again.

PR #201, opened 2020-05-31
SDk-1554: Doc Scan Demo

echarrod:
- Uses .NET Core 3.1
- Put DOCSCAN.md under the `docs` folder, as per the new decided structure. Profile & AML additions will follow in a separate PR, together with links to each of these from the main README, and a new structure.
- Didn't include the same bootstrap static files as we have in the other SDKs, as they are included through the NuGet package (automatically included when creating the project). We can then update in the future, without having to worry about changing the static files. 
- I did an auto-format of all of the example files, which modified the existing examples' formatting as well
- I changed `YotiDocsPathPrefix` to end in a `/` (see f83a566). This is an attempt to prevent problems with concatenating URIs (and losing the v1 part). If two URLs are joined together, one absolute and one relative with "new Uri(absolute, relative)", then the last part of the base path is lost if it doesn't end in a "/". This has caused problems before, but we handle this in the RequestBuilder. But since the generating of the iFrame uses this base URL, and can result in problems, I thought it would be safest to include it in the constant to avoid any future erroneous URIs (as it isn't always obvious that it's cut out the last part of the URI!)


src/Yoti.Auth/Constants/Api.cs
```diff
@@ -9,7 +9,7 @@ public static class Api
         public const string YotiApiPathPrefix = "api/v1";
         public readonly static string DefaultYotiApiUrl = string.Join("/", DefaultYotiHost, YotiApiPathPrefix);
 
-        public const string YotiDocsPathPrefix = "idverify/v1";
+        public const string YotiDocsPathPrefix = "idverify/v1/";
```
davidgrayston, line 12: Should this have a trailing slash? (it differs to _YotiApiPathPrefix_  `api/v1`)

echarrod: I left a note about that in the PR description - the last bullet 👍

PR #203, opened 2020-06-01
SDK-1421: Split READMEs out into separate sections

echarrod:
- under "docs" folder:
  - AML
  - Doc Scan
  - Profile 
Instructions on how to run each example also live with the respective projects in a separate README, linked to from the above READMEs under "docs"

Obviously all the links to other parts have changed, pretty sure I've checked them all, but another click-through would be helpful. 

Please check the wording under "Services" in the main README, i used the text from the developer docs, but I'm open to any improvements.

A lot of the context has been changed from the documentation, so please confirm that they still make sense.


jlmartyres:
LGTM

PR #239, opened 2020-09-21
IN-3116: dotNET Example QR

nikhilPank:


echarrod (changes requested):
We'll need to update the .NET Core Example as well, right?

davidgrayston (changes requested):
Do we need to update the dynamic share example too?

https://github.com/getyoti/yoti-dotnet-sdk/blob/development/src/Examples/Profile/47Example/Views/Home/DynamicScenario.cshtml#L77

echarrod (approved):

davidgrayston (approved):

PR #332, opened 2021-11-03
SDK-1971 Add Face Capture Resource To IDV

BenSmithYoti:
Add Face Capture Resource to IDV with relevant tests
Add test for GeneratedProfileResponse

### DIFF
--- a/src/Yoti.Auth/DocScan/DocScanService.cs
+++ b/src/Yoti.Auth/DocScan/DocScanService.cs
@@ -4,8 +4,11 @@ using System.Text;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Org.BouncyCastle.Crypto;
+using Yoti.Auth.Constants;
 using Yoti.Auth.DocScan.Session.Create;
+using Yoti.Auth.DocScan.Session.Create.FaceCapture;
 using Yoti.Auth.DocScan.Session.Retrieve;
+using Yoti.Auth.DocScan.Session.Retrieve.CreateFaceCaptureResourceResponse;
 using Yoti.Auth.DocScan.Support;
 using Yoti.Auth.Exceptions;
 using Yoti.Auth.Web;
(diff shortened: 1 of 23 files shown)


src/Yoti.Auth/DocScan/DocScanService.cs
```diff
@@ -238,5 +302,7 @@ private static string MediaEndpoint(string sessionId, string mediaId)
         {
             return $"sessions/{sessionId}/media/{mediaId}/content";
         }
+
+        private static JsonSerializerSettings DefaultJsonSettings => new JsonSerializerSettings { NullValueHandling = NullValueHandling.Ignore };
```
echarrod, line 306: Nice, could we call this `YotiDefaultJsonSettings` or something? When I saw it declared above I didn't realise it was internally defined

src/Yoti.Auth/DocScan/DocScanService.cs
```diff
@@ -82,7 +80,7 @@ public async Task<GetSessionResult> GetSession(string sdkId, AsymmetricCipherKey
                 .WithHttpMethod(HttpMethod.Get)
                 .WithBaseUri(ApiUri)
                 .WithEndpoint(sessionEndpoint)
-                .WithQueryParam("sdkId", sdkId)
+                .WithQueryParam("sdkId", sdkId)
```
echarrod, line 83: I think a bit of whitespace has crept in here. 
```suggestion
                .WithQueryParam("sdkId", sdkId)
```
Ideally we'd have a pre-commit step to trim things like this. We can add this at a later stage. 
Something like [this ](https://github.com/dotnet/format/blob/main/docs/integrations.md#pre-commitcom-hook-to-reformat)might work

echarrod (commented):
Nice work Ben, just a couple of suggestions :+1:

src/Yoti.Auth/DocScan/DocScanService.cs
BenSmithYoti: Sure. Will amend right now for you.

src/Yoti.Auth/DocScan/DocScanService.cs
BenSmithYoti: Yes I saw that once PR was up on github. It's annoying as VS doesn't (with my settings anyhow - almost certainly configurable) pick up on these, but GitHub's diff viewer does. VS settings make sense for C# in VS (as that whitespace is obviously irrelevant to the MSIL and/or runtime code), GitHub needs to deal with Python etc so does it's diffs differently. Will see if I can amend my VS settings to match GitHub so we avoid going forwards. Thanks

BenSmithYoti: Will amend with new commit. Thanks

echarrod (approved):

PR #334, opened 2021-11-08
SDK-1979:Add Session Deadline To IDV

BenSmithYoti:
Add Session Deadline To IDV
Modify ClientSessionTokenTtl to make it nullable and so omitted when Relying Business has not set


echarrod (approved):
Looks good. Just wondering whether we should make `ResourcesTtl` nullable as well, but we can address that in a separate PR if we do decide to make that change.

PR #354, opened 2022-03-18
SDK-2055-2072: Add Identity Profile Requirements to DynamicPolicyBuil…

BenSmithYoti:
…der. Add Subject to DynamicScenarioBuilder. Add relevant tests. Modify SignedMessageFactory to remove unncessary (& warning producing) ToString.

### DIFF
--- a/src/Yoti.Auth/Web/SignedMessageFactory.cs
+++ b/src/Yoti.Auth/Web/SignedMessageFactory.cs
@@ -13,7 +13,7 @@ namespace Yoti.Auth.Web
                 throw new ArgumentNullException(nameof(httpMethod));
             }
 
-            string stringToConvert = $"{httpMethod.ToString()}&{endpoint}";
+            string stringToConvert = $"{httpMethod}&{endpoint}";
 
             if (content != null)
                 stringToConvert += "&" + Conversion.BytesToBase64(content);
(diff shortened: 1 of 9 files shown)


src/Yoti.Auth/Web/SignedMessageFactory.cs
```diff
@@ -13,7 +13,7 @@ public static string SignMessage(HttpMethod httpMethod, string endpoint, Asymmet
                 throw new ArgumentNullException(nameof(httpMethod));
             }
 
-            string stringToConvert = $"{httpMethod.ToString()}&{endpoint}";
+            string stringToConvert = $"{httpMethod}&{endpoint}";
```
echarrod, line 16: Presume this was an unnecessary cast?

test/Yoti.Auth.Tests/ShareUrl/Policy/DynamicPolicyBuilderTests.cs
```diff
@@ -312,5 +312,18 @@ public void ShouldBuildWithRememberMeFlag()
 
             Assert.IsTrue(result.WantedRememberMeId);
         }
+
+
```
echarrod, line 316:
```suggestion
```

src/Yoti.Auth/Web/SignedMessageFactory.cs
BenSmithYoti: yes indeed totally unecessary ToString() call inside interpolation, plus was kicking out warning.

laurent-yoti (commented):
Makes sense to me, can not really comment on code style though.

echarrod (commented):
Looks good. Let's wait until we can test before merging though :+1:

echarrod (approved):
Thinking about it, will be a pain to test if we are doing all separate PRs. Happy for this to be merged and then further iterations made if needed

PR #360, opened 2022-03-23
SDK-2056: Add Identity Profile Report attribute

echarrod:
### Added
- Identity Profile Report attribute (type JSON)
  - See [RTWIdentityProfileReport.json](https://github.com/getyoti/yoti-dotnet-sdk/blob/776f4e852cb3ab50fa1251952940b039686fcdd0/test/Yoti.Auth.Tests/TestData/RTWIdentityProfileReport.json) for an example of the JSON value

### DIFF
--- a/src/Yoti.Auth/Profile/YotiProfile.cs
+++ b/src/Yoti.Auth/Profile/YotiProfile.cs
@@ -183,6 +183,19 @@ namespace Yoti.Auth.Profile
         }
 
         /// <summary>
+        /// IdentityProfileReport represents the JSON object containing identity assertion and the
+        /// verification report. This will be null if not provided by Yoti.
+        /// </summary>
+        public YotiAttribute<Dictionary<string, JToken>> IdentityProfileReport
+        {
+            get
+            {
+                return GetAttributeByName<Dictionary<string, JToken>>(name: Constants.UserProfile.IdentityProfileReportAttribute);
+            }
+        }
+
+
+        /// <summary>
         /// Finds all the 'Age Over' and 'Age Under' derived attributes returned with the profile,
         /// and returns them wrapped in <see cref="AgeVerification"/> objects. Returns null if no
         /// matches were found.
--- a/test/Yoti.Auth.Tests/ActivityTests.cs
+++ b/test/Yoti.Auth.Tests/ActivityTests.cs
@@ -584,9 +584,8 @@ namespace Yoti.Auth.Tests
 
             _yotiProfile = TestTools.Profile.CreateUserProfileWithSingleAttribute<Dictionary<string, JToken>>(attribute);
 
-            Dictionary<string, JToken> structuredPostalAddress = _yotiProfile.IdentityProfileReport.GetValue();
-            AssertDictionaryValue("<signature provided here>", "proof", structuredPostalAddress);
-
+            Dictionary<string, JToken> identityProfileReport = _yotiProfile.IdentityProfileReport.GetValue();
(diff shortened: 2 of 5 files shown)


test/Yoti.Auth.Tests/ActivityTests.cs
```diff
@@ -564,6 +565,30 @@ public void DocumentImagesAttributeShouldBeAddedToProfile()
...
+            var attribute = new ProtoBuf.Attribute.Attribute
+            {
+                Name = Constants.UserProfile.IdentityProfileReportAttribute,
+                ContentType = ContentType.Json,
+                Value = ByteString.CopyFromUtf8(json)
+            };
+
+            _yotiProfile = TestTools.Profile.CreateUserProfileWithSingleAttribute<Dictionary<string, JToken>>(attribute);
+
+            Dictionary<string, JToken> structuredPostalAddress = _yotiProfile.IdentityProfileReport.GetValue();
```
MemeDeveloper, line 587: Naming and assertion here don’t look right to me.

BenSmithYoti (approved):

test/Yoti.Auth.Tests/ActivityTests.cs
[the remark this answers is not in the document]
echarrod: Fixed in https://github.com/getyoti/yoti-dotnet-sdk/pull/360/commits/e330aa62acaa494c150433ba0385fa1f65e1b70d

LuborRod (approved):

src/Yoti.Auth/Profile/YotiProfile.cs
```diff
@@ -182,6 +182,19 @@ public YotiAttribute<List<Image>> DocumentImages
             }
         }
 
+        /// <summary>
+        /// IdentityProfileReport represents the JSON object containing identity assertion and the
+        /// verification report. This will be null if not provided by Yoti.
+        /// </summary>
+        public YotiAttribute<Dictionary<string, JToken>> IdentityProfileReport
```
BenSmithYoti, line 189: Ed. Just wondering, in the same vein as discussion on the Java side... might we perhaps be better off here using `YotiAttribute<dynamic> `
as opposed 
`YotiAttribute<Dictionary<string, JToken>>`
...as although the latter should handle the current json schema fine - it will (I believe) break if the IdentityProfileReport was ever changed to an array (as opposed to an object) at it's root. I think this is probably unlikely, but worth mentioning perhaps.

echarrod: @pgreci I know it would be a breaking change, but do you think we ever might change the IdentityProfileReport to have an array, rather than an object with string keys? 
My thinking would be that if that were the case, we'd introduce `identity_profile_reports` (plural), so as not to be a breaking change. But just wanted to confirm.

laurent-yoti: If I understand well, isn't it the same case as for the `document_images` which could now appear several time and for which we need to introduce a way of managing all its occurrence by introducing something like `getAttributeByNameAndId()`.... if so, the same principle could be applied to the attribute `identity_profile_report` and in case there are several of them, we'd be able to find the right one by specifying an extra id  ?

echarrod: Yep, think we just get the right one by ID, I don't think we need `name` as well, assuming the IDs are unique? So `getAttributeById`?

BenSmithYoti (approved):

PR #122, opened 2019-10-31
SDK-1150: Same-named Attributes

echarrod:
### Added
- `GetAttributesByName`:
```c#
var selfieAttributes = profile.GetAttributesByName<JpegImage>(name: "selfie");
var documentDetailsAttributes = profile.GetAttributesByName<DocumentDetails>(name: "document_details");
var documentImagesAttributes = profile.GetAttributesByName<List<Image>>(name: "document_images");
```

### Deprecated 
- `yotiProfile.Attributes`.
  -  Use `yotiProfile.AttributeCollection` instead

Tested that 2 x Passport Details are retrieved by the SDK and shown in the Core Example project

PR #362, opened 2022-03-25
SDK-1150: Fix FindAttributesStartingWith for Derived Attributes 

echarrod:
### Fixed 
- Update FindAttributesStartingWith to loop through each attribute with the same name, after AttributeCollection was added in https://github.com/getyoti/yoti-dotnet-sdk/pull/122

We can't unit test this for this specific scenario as the logic is from the Age Over/Under attribute being derived before the attribute list is shared. This is covered in integration tests though


BenSmithYoti (approved):
Looks good.

PR #366, opened 2022-03-29
NA: Fix Sonar code smell

echarrod:
### Changed
- Missed a code smell in https://github.com/getyoti/yoti-dotnet-sdk/pull/362

"SonarQube: Loops should be simplified with "LINQ" expressions"
> When a loop is filtering, selecting or aggregating, those functions can be handled with a clearer, more concise LINQ expression instead.

I've run the integration tests against this branch to check it doesn't modify behaviour

### DIFF
--- a/src/Yoti.Auth/Profile/BaseProfile.cs
+++ b/src/Yoti.Auth/Profile/BaseProfile.cs
@@ -150,14 +150,10 @@ namespace Yoti.Auth.Profile
         {
             if (prefix == null)
                 throw new ArgumentNullException(nameof(prefix));
-            return 
-                _attributes.Where(
-                    attributesByName => attributesByName.Key.StartsWith(
-                        prefix, StringComparison.Ordinal))
-                .SelectMany(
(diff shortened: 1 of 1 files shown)


src/Yoti.Auth/Profile/BaseProfile.cs
```diff
@@ -150,24 +150,14 @@ public List<YotiAttribute<T>> FindAttributesStartingWith<T>(string prefix)
...
-                    }
-                }
-            }
-
-            return matches;
+            return 
+                _attributes.Where(
+                    attributesByName => attributesByName.Key.StartsWith(
+                        prefix, StringComparison.Ordinal))
+                .SelectMany(
```
BenSmithYoti, line 157: Seems condition 
`(attribute is YotiAttribute<T> castableAttribute)`
is not implemented here. 
If this changes the behaviour (?) we could fix with a suitable 
`.Where(a => a is YotiAttribute<T>)`

BenSmithYoti (approved):
Look good.

PR #367, opened 2022-03-31
SDK-2056: Add Identity Profile Report attribute to example project

echarrod:
### Added
- Identity Profile Report attribute to example project


BenSmithYoti (approved):
Look good.

PR #380, opened 2022-05-19
SDK-2140: Add ErrorCode and ResponseContent to YotiProfileException

BenSmithYoti:
to surface error details for failure receipts for Digital ID App Shares

### DIFF
--- a/src/Yoti.Auth/Exceptions/YotiProfileException.cs
+++ b/src/Yoti.Auth/Exceptions/YotiProfileException.cs
@@ -1,22 +1,35 @@
-﻿using System;
+﻿using Newtonsoft.Json.Linq;
+using System;
 
 namespace Yoti.Auth.Exceptions
 {
-    public class YotiProfileException : YotiException
-    {
-        public YotiProfileException()
-            : base()
-        {
-        }
+	public class YotiProfileException : YotiException
+	{
+		public YotiProfileException()
+			: base()
+		{
+		}
 
-        public YotiProfileException(string message)
-            : base(message)
-        {
-        }
+		public YotiProfileException(string message)
+			: base(message)
+		{
+		}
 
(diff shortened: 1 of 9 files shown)


src/Yoti.Auth/Exceptions/YotiProfileException.cs
```diff
@@ -1,22 +1,35 @@
...
-            : base(message, innerException)
-        {
-        }
-    }
+		public YotiProfileException(string message, string responseContent)
+		 : base(message)
+		{
+			ResponseContent = responseContent;
+			dynamic jsonResponse = JObject.Parse(responseContent);
+			if (jsonResponse.error_details != null && jsonResponse.error_details.error_code != null)
```
echarrod, line 23: I think we want to include the Error Description in the exception too Ben, as it contains useful information

BenSmithYoti: It is included in the ResponseContent.

BenSmithYoti: I could create another prop on the exception (but not today) for description, but I understood the error_code was the vital thing PG wanted surfaced, but tht all the response should also be available - so that's what I dev'd against. Java went a different way but I wanted to be simple, and not bind the SDK too tightly to what connect have done with this version of API, so buffering itnegrators from changes.

BenSmithYoti: Happy to ammend if that's what is needed, but unlikely to have chance until June 13th of later.

echarrod: Ah yes I guess it's available in the response. I think it would be good, but I don't feel too strongly either way to be honest. We won't be able to release this until after June 13th anyway

echarrod (approved):

PR #381, opened 2022-05-20
SDK-2140: Minor Fixes to Error Code Surfacing

BenSmithYoti:
YotiProfileException Message no longer breaks onto two lines
Mods to fix minor code smells from previous [PR](https://github.com/getyoti/yoti-dotnet-sdk/pull/380)

### DIFF
--- a/src/Yoti.Auth/Yoti.Auth.csproj
+++ b/src/Yoti.Auth/Yoti.Auth.csproj
@@ -16,10 +16,10 @@
     <PackageIcon>Yoti.png</PackageIcon>
     <RepositoryUrl>https://github.com/getyoti/yoti-dotnet-sdk</RepositoryUrl>
     <PackageTags>yoti 2FA multifactor authentication verification identity login register verify 2Factor sdk sandbox documents docs KYC AML</PackageTags>
-    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
     <LangVersion>latest</LangVersion>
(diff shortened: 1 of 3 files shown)


BenSmithYoti:
N.B. lots of whitespace changes, so ignore whitespace for diffs.

src/Yoti.Auth/Yoti.Auth.csproj
```diff
@@ -16,10 +16,10 @@
     <PackageIcon>Yoti.png</PackageIcon>
     <RepositoryUrl>https://github.com/getyoti/yoti-dotnet-sdk</RepositoryUrl>
     <PackageTags>yoti 2FA multifactor authentication verification identity login register verify 2Factor sdk sandbox documents docs KYC AML</PackageTags>
-    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
     <LangVersion>latest</LangVersion>
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
-    <Version>3.11.0</Version>
+    <Version>3.12.1-beta</Version>
```
echarrod, line 22: Would prefer not to have this as the version in development :+1: we should update when we can

BenSmithYoti: 100% agreed, I only left it like this so that in both our absences (holidays) there was no confusion about the exact codebase the package I delivered to the slack channel was generated from. Can ammend in next PR. Thanks Ed.

PR #439, opened 2023-10-06
Sdk 2245 net retrieve session

mehmet-yoti:


saurabh-yoti (changes requested):
Need to swap fields for `GetSessionResult` and `ShareSessionResult` classes. 

When a Session is created (ShareSessionResult), these properties should be returned:
`id, status, expiry`

When a Session is fetched (GetSessionResult), these properties should be returned:
`id, status, expiry, created, updated, qrCode, receipt`

src/Yoti.Auth/DigitalIdentity/GetSessionResult.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        [JsonProperty("status")]
+        public string Status { get; set; }
+
+        [JsonProperty("expiry")]
+        public string Expiry { get; set; }
+
+#pragma warning restore 0649
+
+    }
```
irotech, line 21: Please refer to our internal documentation (sent link in DM)
Depending on the session status you can find a subset of the properties of COMPLETED Session.

If you select the example "Session with status COMPLETED" you will find all the properties we need to map

src/Yoti.Auth/DigitalIdentity/DigitalIdentityService.cs
```diff
@@ -51,5 +51,43 @@ internal static async Task<ShareSessionResult> CreateShareSession(HttpClient htt
...
+
+                return deserialized;
+
+            }
+        }
+
     }
+
+
 }
```
irotech, line 93: In the Java SDK we have a new line as EOF for all classes as convention. Please ignore if it does not apply in here.

irotech (changes requested):

saurabh-yoti (approved):

src/Yoti.Auth/DigitalIdentity/GetSessionResult.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    public class qrCode
+    {
+        [JsonProperty("id")]
+        public string Id { get; set; }
+        [JsonProperty("uri")]
+        public string Uri { get; set; }
```
irotech, line 40: minor: Even if it doesn't hurt mapping the `qrCode` URI may lead to confusion when an integrator finds a getter/setter in the API and no use for it.

Additionally the SDKs should be aligned as much as possible [[Java](https://github.com/getyoti/yoti-java-sdk/blob/master/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSession.java), [Node](https://github.com/getyoti/yoti-node-sdk/blob/master/src/digital_identity_service/get.share.session.result.js)]

src/Yoti.Auth/DigitalIdentity/DigitalIdentityService.cs
irotech: minor: Is there a shared coding style in .NET?

irotech (approved):
Up to you if you want to address the minor suggestions

src/Yoti.Auth/DigitalIdentity/ShareSessionResult.cs
```diff
@@ -1,14 +1,11 @@
-﻿using System;
-using System.Reflection.Emit;
-using Newtonsoft.Json;
-using Yoti.Auth.DataObjects;
+﻿using Newtonsoft.Json;
+
 
 namespace Yoti.Auth.DigitalIdentity
 {
     public class ShareSessionResult
```
irotech, line 6: `status` is missing

irotech (changes requested):
sorry I thouth there was a missing property in the session creation response obj but it was just not present in the diff

irotech (approved):

src/Yoti.Auth/DigitalIdentity/GetSessionResult.cs
mehmet-yoti: You are right i have removed Uri to be align all sdk.

src/Yoti.Auth/DigitalIdentity/ShareSessionResult.cs
mehmet-yoti: I have checked and status is in ShareSessionResult file.  So it seems there is no problem

PR #444, opened 2023-12-15
Sdk 2368 net add support for advanced identity profiles to share v 1 v 2 and examples

mehmet-yoti:
- Added advanced Identity Profile for Share and Idv


irotech (approved):
It needs a rebase as it is in conflict with the base

PR #447, opened 2024-01-18
SDK-2341: added support to retrieve expanded doc fields and media

mehmet-yoti:
SDK-2341: added support to retrieve expanded doc fields and media, added success page example


saurabh-yoti (approved):
Looks good to me!

PR #538, opened 2026-04-19

mehmet-yoti:
## Summary
Adds support for the IDV shortened flow by allowing integrators to configure which IDV client screens should be suppressed (omitted) during a session. A new `suppressed_screens` field is exposed on `SdkConfig`, together with builder methods and a fixed set of valid screen identifiers defined as constants.

## Changes
- `src/Yoti.Auth/Constants/DocScanConstants.cs`: Added seven new screen identifier constants (`ID_DOCUMENT_EDUCATION`, `ID_DOCUMENT_REQUIREMENTS`, `SUPPLEMENTARY_DOCUMENT_EDUCATION`, `ZOOM_LIVENESS_EDUCATION`, `STATIC_LIVENESS_EDUCATION`, `FACE_CAPTURE_EDUCATION`, `FLOW_COMPLETION`).
- `src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs`: Added a `SuppressedScreens` property (serialized as `suppressed_screens` with `NullValueHandling.Ignore`) and a new optional constructor parameter.
- `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs`: Added `WithSuppressedScreen(string)` and `WithSuppressedScreens(List<string>)` builder methods. Invalid/unknown identifiers are filtered out silently, duplicates are removed, and passing `null` to `WithSuppressedScreens` clears previously configured values.
- `test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs`: Added 16 unit tests covering default state, single/multiple values, filtering of unknown/null/empty identifiers, deduplication, replacement/clearing semantics, and JSON (de)serialization.

## QA Test Steps
1. **Setup**
   - Check out branch `websdk-auto/SDK-2612-dotnet-idv---support-configuration-for-idv-shortened-flow`.
   - Restore and build: `dotnet restore && dotnet build`.
2. **Run unit tests**
   - `dotnet test test/Yoti.Auth.Tests --filter "FullyQualifiedName~SdkConfigBuilderTests"`.
   - Verify all new `SuppressedScreens*` tests pass alongside existing tests (no regressions).
3. **Happy path — single screen**
   - Build an `SdkConfig` with `.WithSuppressedScreen(DocScanConstants.IdDocumentEducation).Build()`.
   - Serialize via `JsonConvert.SerializeObject` and confirm JSON contains `"suppressed_screens": ["ID_DOCUMENT_EDUCATION"]`.
4. **Happy path — all valid screens**
   - Chain `.WithSuppressedScreen(...)` for all seven constants; assert `SuppressedScreens.Count == 7`.
5. **List API**
   - Call `.WithSuppressedScreens(new List<string> { IdDocumentEducation, FlowCompletion })`; assert both appear in the built config.
6. **Edge cases**
   - Unknown identifier (e.g. `"FOO_BAR"`): confirm it is silently ignored and `SuppressedScreens` remains `null` if no other values set.
   - Null/empty string passed to `WithSuppressedScreen`: confirm it is ignored.
   - Duplicate values: add the same identifier twice; confirm only one entry exists.
   - `WithSuppressedScreens(null)` after adding values: confirm the list is cleared (`SuppressedScreens == null`).
   - `WithSuppressedScreens(new List<string>{…})` after prior single adds: confirm the list replaces previous values (not merged).
   - Mixed list with unknown values: confirm unknowns filtered, valid ones retained.
7. **Serialization regression checks**
   - Build an `SdkConfig` with no suppressed screens; confirm serialized JSON does NOT contain the `suppressed_screens` key.
   - Deserialize `{}` → `SuppressedScreens` should be `null`.
   - Deserialize `{ "suppressed_screens": null }` → `null`.
   - Deserialize `{ "suppressed_screens": [] }` → empty list.
   - Deserialize `{ "suppressed_screens": ["ID_DOCUMENT_EDUCATION","FLOW_COMPLETION"] }` → list of two values preserved.
8. **End-to-end (optional, if IDV backend available)**
   - Create a DocScan session with a `SessionSpec` containing the new `SdkConfig` with one or more suppressed screens; confirm the request payload sent to the IDV API includes `suppressed_screens` and that the resulting IDV web client flow skips the suppressed screen(s).
9. **Regression**
   - Re-run existing `SdkConfigBuilderTests` (capture methods, colours, URLs, attempts configuration) to confirm none are affected by the new constructor parameter.

## Notes
- Invalid/unknown screen identifiers are silently filtered rather than raising an exception. This is intentional so the SDK remains forward-compatible if new screen IDs are introduced server-side before the SDK is updated — see the XML doc on `WithSuppressedScreen`.
- The `suppressed_screens` JSON field uses `NullValueHandling.Ignore`, so it is only emitted when at least one screen has been configured; this avoids changing the request payload for existing integrators.
- `WithSuppressedScreens(List<string>)` *replaces* any previously configured values rather than merging, matching the typical builder "setter" pattern. Passing `null` clears the list entirely.
- Valid screen identifiers are stored in a private `HashSet<string>` within the builder; keep this set in sync with `DocScanConstants` if additional screens are added in the future.

---

*Related Jira:* SDK-2612

### DIFF
--- a/src/Yoti.Auth/Constants/DocScanConstants.cs
+++ b/src/Yoti.Auth/Constants/DocScanConstants.cs
@@ -74,5 +74,13 @@
         public const string Generic = "GENERIC";
 
         public const string VerifyShareCodeTask = "VERIFY_SHARE_CODE_TASK";
+
+        public const string IdDocumentEducation = "ID_DOCUMENT_EDUCATION";
+        public const string IdDocumentRequirements = "ID_DOCUMENT_REQUIREMENTS";
+        public const string SupplementaryDocumentEducation = "SUPPLEMENTARY_DOCUMENT_EDUCATION";
+        public const string ZoomLivenessEducation = "ZOOM_LIVENESS_EDUCATION";
+        public const string StaticLivenessEducation = "STATIC_LIVENESS_EDUCATION";
+        public const string FaceCaptureEducation = "FACE_CAPTURE_EDUCATION";
+        public const string FlowCompletion = "FLOW_COMPLETION";
     }
 }
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
@@ -1,4 +1,4 @@
-﻿using Newtonsoft.Json;
+using Newtonsoft.Json;
 using System.Collections.Generic;
 
 namespace Yoti.Auth.DocScan.Session.Create
@@ -41,6 +41,9 @@ namespace Yoti.Auth.DocScan.Session.Create
         [JsonProperty(PropertyName = "attempts_configuration")]
         public AttemptsConfiguration AttemptsConfiguration { get; }
 
+        [JsonProperty(PropertyName = "suppressed_screens", NullValueHandling = NullValueHandling.Ignore)]
+        public List<string> SuppressedScreens { get; }
+
         public SdkConfig(string allowedCaptureMethods,
                             string primaryColour,
                             string secondaryColour,
@@ -51,7 +54,8 @@ namespace Yoti.Auth.DocScan.Session.Create
                             string errorUrl,
                             string privacyPolicyUrl,
                             bool? allowHandoff = null,
-                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null)
+                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null,
+                            List<string> suppressedScreens = null)
         {
             AllowedCaptureMethods = allowedCaptureMethods;
             PrimaryColour = primaryColour;
@@ -71,6 +75,8 @@ namespace Yoti.Auth.DocScan.Session.Create
                     IdDocumentTextDataExtraction = idDocumentTextDataExtractionRetriesConfig
                 };
             }
+
+            SuppressedScreens = suppressedScreens;
         }
     }
 }
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
@@ -16,6 +16,18 @@ namespace Yoti.Auth.DocScan.Session.Create
         private string _privacyPolicyUrl;
         private bool? _allowHandoff;
         private Dictionary<string, int> _idDocumentTextDataExtractionAttemptsConfig;
+        private List<string> _suppressedScreens;
+
+        private static readonly HashSet<string> _validSuppressedScreens = new HashSet<string>
+        {
+            DocScanConstants.IdDocumentEducation,
+            DocScanConstants.IdDocumentRequirements,
+            DocScanConstants.SupplementaryDocumentEducation,
+            DocScanConstants.ZoomLivenessEducation,
+            DocScanConstants.StaticLivenessEducation,
+            DocScanConstants.FaceCaptureEducation,
+            DocScanConstants.FlowCompletion
+        };
 
         /// <summary>
         /// Sets the allowed capture method to "CAMERA"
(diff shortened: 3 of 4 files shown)


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Critical
None

### Major
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:261-273` — Silent filtering of unknown screen identifiers is a risky API design.** The builder silently drops unknown/null/empty values, so callers get no signal that the screen they wanted to suppress was rejected. The justification in the doc-comment ("so that new backend values don't cause failures") is inverted: this code runs in the *caller's* SDK, which already has a fixed allow-list pinned to the SDK version, so the "new backend values" case cannot actually arrive here. The real effect is that a typo (e.g. `"ID_DOCUEMENT_EDUCATION"`) will ship to production and silently do nothing. This also locks backend/SDK evolution together: any new screen added server-side requires a client SDK release before integrators can use it. Suggested fix: either (a) throw `ArgumentException` for unknown/null/empty values, matching the fail-fast style used elsewhere in this SDK's builders, or (b) drop the allow-list entirely and forward whatever the caller provides (the server is the source of truth). If the allow-list is kept for ergonomics, log a warning or at minimum document that invalid values are silently ignored.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:288-302` — `WithSuppressedScreens` has surprising replacement semantics that will confuse callers combining the two setters.** `WithSuppressedScreen(A).WithSuppressedScreens(new List<string>{ B })` produces `[B]`, but `WithSuppressedScreens(new List<string>{ B }).WithSuppressedScreen(A)` produces `[B, A]`. That ordering-sensitive behavior is easy to trip over and is not obvious from the method name. It's also inconsistent with the builder's own `WithIdDocumentTextExtractionCategoryAttempts` pattern (additive). Suggested fix: pick one semantic and document it clearly in the XML doc. Either rename to `SetSuppressedScreens` to signal replacement, or make it additive like the rest of the builder (`foreach WithSuppressedScreen`) without the `_suppressedScreens = null` reset on line 296.

### Minor
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:290-294` — Passing `null` to `WithSuppressedScreens` clears state, which is an undocumented side effect that diverges from other setters.** No other `With...` setter in this builder treats `null` as "reset" (e.g. `WithPrimaryColour(null)` stores `null` as the value). Suggested fix: if intentional, expose an explicit sibling method `ClearSuppressedScreens()` / `ResetSuppressedScreens()`, and let `WithSuppressedScreens(null)` throw `ArgumentNullException` for consistency.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs:44-45` — `SuppressedScreens` is exposed as a mutable `List<string>` on an otherwise immutable value object.** Callers can mutate the list post-build (`sdkConfig.SuppressedScreens.Add(...)`), undermining immutability and allowing two sessions built from the same builder to diverge. Suggested fix: either defensively copy on assignment (`SuppressedScreens = suppressedScreens != null ? new List<string>(suppressedScreens) : null;`), or expose the property type as `IReadOnlyList<string>` / `IReadOnlyCollection<string>`.
- **`test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs:455-463` — Missing test for the interaction path between `WithSuppressedScreen` and `WithSuppressedScreens`.** The filter path and the happy path are covered, but the ordering-sensitivity called out above is not. Suggested fix: add two tests asserting the exact ordering semantics (additive vs. replacement) so the behavior becomes part of the contract.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:21-30` — `_validSuppressedScreens` uses default (ordinal-case-sensitive) `HashSet<string>` equality, but case handling is not asserted anywhere.** A caller accidentally passing `"id_document_education"` will be silently dropped. Suggested fix: either add an explicit `StringComparer.Ordinal` (documenting intent) plus a test covering case-mismatch rejection, or use `StringComparer.OrdinalIgnoreCase` if the server is case-insensitive.

### Nit
- **`src/Yoti.Auth/Constants/DocScanConstants.cs:86` — File is missing a trailing newline.** `SdkConfig.cs` was corrected to add one in this diff — worth doing the same here.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs:1` — UTF-8 BOM was stripped from this file in the diff.** Other files in the SDK (including `SdkConfigBuilderTests.cs`) still have the BOM. Harmless, but it creates noise in future diffs/blame. Suggested fix: keep BOM usage consistent across the project.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:21` — `_validSuppressedScreens` is a static member but uses the instance-field underscore prefix.** Suggested fix: rename to `ValidSuppressedScreens` (PascalCase for non-private statics) or match whichever static-field convention is used elsewhere.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:249` — Pre-existing trailing whitespace was replaced, introducing a noise-only line change.** Consider a pre-commit hook to normalize so whitespace diffs don't bundle with feature diffs.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:261` — Method name `WithSuppressedScreen` (singular) vs. `WithSuppressedScreens` (plural) is easy to misread.** Consider `AddSuppressedScreen` for the singular, which better describes the additive semantics and matches the common builder idiom (`Add...` for append, `With...`/`Set...` for replace).

PR #539, opened 2026-04-20

mehmet-yoti:
## Summary
Adds support for configuring the IDV shortened flow by allowing clients to suppress specific screens in the document scan session SDK configuration. This introduces a new `WithSuppressedScreens` builder method on `SdkConfigBuilder` that accepts a list of screen identifiers, which are serialized as `suppressed_screens` in the session creation payload and omitted when not set.

## Changes
- `src/Yoti.Auth/Constants/DocScanConstants.cs`: Added seven new screen identifier constants (`ID_DOCUMENT_EDUCATION`, `ID_DOCUMENT_REQUIREMENTS`, `SUPPLEMENTARY_DOCUMENT_EDUCATION`, `ZOOM_LIVENESS_EDUCATION`, `STATIC_LIVENESS_EDUCATION`, `FACE_CAPTURE_EDUCATION`, `FLOW_COMPLETION`).
- `src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs`: Added `SuppressedScreens` property serialized as `suppressed_screens` with `NullValueHandling.Ignore`; extended constructor with an optional `suppressedScreens` parameter.
- `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs`: Added private `_suppressedScreens` field and public `WithSuppressedScreens(List<string>)` method; wired it into `Build()`.
- `test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs`: Added 9 tests covering builder behavior, constant values, serialization (present/omitted), and deserialization (present/missing/empty/unknown values).

## QA Test Steps
1. **Setup**: Check out the branch and run `dotnet restore` and `dotnet build` for the solution.
2. **Run unit tests**: Execute `dotnet test test/Yoti.Auth.Tests/Yoti.Auth.Tests.csproj --filter "FullyQualifiedName~SdkConfigBuilderTests"` and confirm all tests pass — pay particular attention to the nine new tests:
   - `ShouldBuildWithSuppressedScreens`
   - `SuppressedScreensShouldBeNullIfNotSet`
   - `SuppressedScreenConstantsShouldMatchExpectedValues`
   - `SerializedSdkConfigShouldIncludeSuppressedScreensWhenSet`
   - `SerializedSdkConfigShouldOmitSuppressedScreensWhenNotSet`
   - `DeserializedSdkConfigShouldParseSuppressedScreens`
   - `DeserializedSdkConfigShouldHandleMissingSuppressedScreens`
   - `DeserializedSdkConfigShouldHandleEmptySuppressedScreensArray`
   - `DeserializedSdkConfigShouldHandleUnknownSuppressedScreenValue`
3. **Happy path — builder integration**: In a sample app, build an `SdkConfig` via `new SdkConfigBuilder().WithSuppressedScreens(new List<string> { DocScanConstants.IdDocumentEducation, DocScanConstants.FlowCompletion }).Build();` and assert `SuppressedScreens` contains both values in order.
4. **Happy path — JSON payload**: Serialize the above `SdkConfig` with `JsonConvert.SerializeObject(...)` and confirm the output contains `"suppressed_screens":["ID_DOCUMENT_EDUCATION","FLOW_COMPLETION"]`.
5. **Omission behavior**: Build an `SdkConfig` without calling `WithSuppressedScreens`, serialize it, and confirm the JSON does **not** contain the `suppressed_screens` key (validates `NullValueHandling.Ignore`).
6. **Empty list**: Call `.WithSuppressedScreens(new List<string>())` and confirm it serializes as `"suppressed_screens":[]` (not omitted).
7. **End-to-end regression**: Create a DocScan session against a test tenant using an `SdkConfig` built with suppressed screens and confirm the backend accepts the payload and (if supported in the test environment) the IDV web flow skips the suppressed screens.
8. **Backwards-compatibility regression**: Re-run any existing DocScan session-creation smoke tests that do not use `WithSuppressedScreens` to confirm existing consumers are unaffected.
9. **Deserialization regression**: Deserialize a session config response containing `suppressed_screens` and confirm the list is populated on the resulting `SdkConfig` object.

## Notes
- Screen identifiers are modeled as `List<string>` plus string constants rather than a strongly-typed enum, matching the existing pattern used elsewhere in `DocScanConstants` (e.g., `Generic`, `VerifyShareCodeTask`). This keeps the SDK forward-compatible if the backend adds new screen types before the SDK is updated — validated by the `DeserializedSdkConfigShouldHandleUnknownSuppressedScreenValue` test.
- No validation is performed on the contents of the list; the caller is trusted to pass valid values (documented in the XML docs on `WithSuppressedScreens`).
- `NullValueHandling.Ignore` is applied so that existing session-creation payloads remain byte-identical when the new field is not used — important for consumers who diff or snapshot request bodies.
- The diff also normalizes BOMs/trailing newlines in `SdkConfig.cs` and `SdkConfigBuilder.cs`; these are cosmetic and non-functional.

---

*Related Jira:* SDK-2612

### DIFF
--- a/src/Yoti.Auth/Constants/DocScanConstants.cs
+++ b/src/Yoti.Auth/Constants/DocScanConstants.cs
@@ -74,5 +74,13 @@
         public const string Generic = "GENERIC";
 
         public const string VerifyShareCodeTask = "VERIFY_SHARE_CODE_TASK";
+
+        public const string IdDocumentEducation = "ID_DOCUMENT_EDUCATION";
+        public const string IdDocumentRequirements = "ID_DOCUMENT_REQUIREMENTS";
+        public const string SupplementaryDocumentEducation = "SUPPLEMENTARY_DOCUMENT_EDUCATION";
+        public const string ZoomLivenessEducation = "ZOOM_LIVENESS_EDUCATION";
+        public const string StaticLivenessEducation = "STATIC_LIVENESS_EDUCATION";
+        public const string FaceCaptureEducation = "FACE_CAPTURE_EDUCATION";
+        public const string FlowCompletion = "FLOW_COMPLETION";
     }
 }
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
@@ -1,4 +1,4 @@
-﻿using Newtonsoft.Json;
+using Newtonsoft.Json;
 using System.Collections.Generic;
 
 namespace Yoti.Auth.DocScan.Session.Create
@@ -41,6 +41,9 @@ namespace Yoti.Auth.DocScan.Session.Create
         [JsonProperty(PropertyName = "attempts_configuration")]
         public AttemptsConfiguration AttemptsConfiguration { get; }
 
+        [JsonProperty(PropertyName = "suppressed_screens", NullValueHandling = NullValueHandling.Ignore)]
+        public List<string> SuppressedScreens { get; }
+
         public SdkConfig(string allowedCaptureMethods,
                             string primaryColour,
                             string secondaryColour,
@@ -51,7 +54,8 @@ namespace Yoti.Auth.DocScan.Session.Create
                             string errorUrl,
                             string privacyPolicyUrl,
                             bool? allowHandoff = null,
-                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null)
+                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null,
+                            List<string> suppressedScreens = null)
         {
             AllowedCaptureMethods = allowedCaptureMethods;
             PrimaryColour = primaryColour;
@@ -63,6 +67,7 @@ namespace Yoti.Auth.DocScan.Session.Create
             ErrorUrl = errorUrl;
             PrivacyPolicyUrl = privacyPolicyUrl;
             AllowHandoff = allowHandoff;
+            SuppressedScreens = suppressedScreens;
 
             if (idDocumentTextDataExtractionRetriesConfig != null)
             {
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
@@ -16,6 +16,7 @@ namespace Yoti.Auth.DocScan.Session.Create
         private string _privacyPolicyUrl;
         private bool? _allowHandoff;
         private Dictionary<string, int> _idDocumentTextDataExtractionAttemptsConfig;
+        private List<string> _suppressedScreens;
 
         /// <summary>
         /// Sets the allowed capture method to "CAMERA"
@@ -234,7 +235,26 @@ namespace Yoti.Auth.DocScan.Session.Create
         {
             WithIdDocumentTextExtractionCategoryAttempts(DocScanConstants.Generic, genericAttempts);
             return this;
-        }   
+        }
+
+        /// <summary>
+        /// Sets the list of screens to suppress from the IDV flow
+        /// </summary>
+        /// <remarks>
+        ///     <para>
+        ///         Valid values are defined as constants on <see cref="DocScanConstants"/>:
+        ///         <c>ID_DOCUMENT_EDUCATION</c>, <c>ID_DOCUMENT_REQUIREMENTS</c>,
+        ///         <c>SUPPLEMENTARY_DOCUMENT_EDUCATION</c>, <c>ZOOM_LIVENESS_EDUCATION</c>,
+        ///         <c>STATIC_LIVENESS_EDUCATION</c>, <c>FACE_CAPTURE_EDUCATION</c>, <c>FLOW_COMPLETION</c>.
+        ///     </para>
+        /// </remarks>
+        /// <param name="suppressedScreens">The list of screens to suppress</param>
+        /// <returns>The <see cref="SdkConfigBuilder"/></returns>
+        public SdkConfigBuilder WithSuppressedScreens(List<string> suppressedScreens)
+        {
+            _suppressedScreens = suppressedScreens;
+            return this;
+        }
 
         /// <summary>
         /// Builds the <see cref="SdkConfig"/> based on values supplied to the builder
@@ -253,7 +273,8 @@ namespace Yoti.Auth.DocScan.Session.Create
                 _errorUrl,
                 _privacyPolicyUrl,
                 _allowHandoff,
-                _idDocumentTextDataExtractionAttemptsConfig);
+                _idDocumentTextDataExtractionAttemptsConfig,
+                _suppressedScreens);
         }
     }
 }
--- a/test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs
+++ b/test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
 using System.Collections.Generic;
 using Yoti.Auth.Constants;
 using Yoti.Auth.DocScan.Session.Create;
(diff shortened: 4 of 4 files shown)


mehmet-yoti:
## 🤖 Claude Code Review

VlhObElIUm9aU0JqYjJSbExYSmxkbWxsZDJWeUlHRm5aVzUwSUhSdklISmxkbWxsZHlCMGFHVWdZMmhoYm1kbGN5QnZiaUIwYUdseklHSnlZVzVqYUNCMmN5QnZjbWxuYVc0dmJXRnpkR1Z5TGdvS1EyOXVkR1Y0ZERvS0xTQkthWEpoT2lCVFJFc3RNall4TWdvdElGTkVTem9nWkc5MGJtVjBDZ3BHYVhKemRDQnlkVzQ2SUdkcGRDQmthV1ptSUc5eWFXZHBiaTl0WVhOMFpYSXVMaTVJUlVGRUlIUnZJSE5sWlNCaGJHd2dZMmhoYm1kbGN5NEtWR2hsYmlCcGJuTndaV04wSUhSb1pTQnRiMlJwWm1sbFpDQm1hV3hsY3lCbWIzSWdZMjl1ZEdWNGRDNEtDazkxZEhCMWRDQmhJRzFoY210a2IzZHVJR052WkdVZ2NtVjJhV1YzSUhkcGRHZzZDaTBnUm1sdVpHbHVaM01nWjNKdmRYQmxaQ0JpZVNCelpYWmxjbWwwZVRvZ1EzSnBkR2xqWVd3Z0x5Qk5ZV3B2Y2lBdklFMXBibTl5SUM4Z1RtbDBDaTBnUldGamFDQm1hVzVrYVc1bklHbHVZMngxWkdWeklHWnBiR1U2YkdsdVpTQnlaV1psY21WdVkyVnpJR0Z1WkNCaElITjFaMmRsYzNSbFpDQm1hWGdLTFNCSlppQnVieUJwYzNOMVpYTWdhVzRnWVNCelpYWmxjbWwwZVN3Z2MyRjVJQ0pPYjI1bElnb0tUM1YwY0hWMElFOU9URmtnYldGeWEyUnZkMjRnYzNSaGNuUnBibWNnZDJsMGFDQWlJeU1nUTI5a1pTQlNaWFpwWlhjZ1JtbHVaR2x1WjNNaUlDMGdibThnY0hKbFlXMWliR1VzSUc1dklHMWxkR0VnWTI5dGJXVnVkR0Z5ZVM0PQo=
Error: Input must be provided either through stdin or as a prompt argument when using --print

mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

**Scope reviewed:** Changes on `websdk-auto/SDK-2612-dotnet-idv---support-configuration-for-idv-shortened-flow` vs `origin/master` (commit `84efe51`). The branch adds a `suppressed_screens` field to `SdkConfig` for the IDV shortened-flow feature, along with builder support, 7 new screen-name constants, and 8 unit tests.

### Critical
None.

### Major
None.

### Minor

**1. `WithSuppressedScreens` stores the caller's list reference directly — external mutation can change the built config**
- **File:** `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:253-257`
- **Issue:** `_suppressedScreens = suppressedScreens;` keeps the caller's reference. A later mutation of the passed-in list (e.g. `list.Add(...)` after `.Build()`) will silently mutate the `SdkConfig.SuppressedScreens` property that's about to be serialized on the wire. Because `SuppressedScreens` is a read-only property exposing a live `List<string>`, downstream code can also mutate it post-build.
- **Why it matters:** Leaky encapsulation on an immutable-looking DTO; makes serialisation non-deterministic under concurrent/shared-list usage. Other builder methods in this same class suffer similar issues (e.g. `_idDocumentTextDataExtractionAttemptsConfig`), so this is consistent with existing patterns — but new code is a good opportunity to stop the bleed.
- **Suggested fix:** Copy on assignment and/or expose as `IReadOnlyList<string>`:
  ```csharp
  public SdkConfigBuilder WithSuppressedScreens(List<string> suppressedScreens)
  {
      _suppressedScreens = suppressedScreens == null
          ? null
          : new List<string>(suppressedScreens);
      return this;
  }
  ```
  If changing the public property type is out of scope for consistency reasons, at least defensively copy in the builder.

**2. No validation or append-style accumulation — second call overwrites the first**
- **File:** `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:253-257`
- **Issue:** Calling `WithSuppressedScreens(a).WithSuppressedScreens(b)` silently discards `a`. Compare with `WithIdDocumentTextExtractionCategoryAttempts` (lines 174–184) which accumulates. Passing `null` also wipes previously set screens.
- **Why it matters:** Fluent-builder ergonomics: the other collection-setter in this class *adds*, so callers may reasonably expect the same shape. At minimum this should be documented.
- **Suggested fix:** Either (a) add an `AddSuppressedScreen(string screen)` overload that lazily initialises and appends, mirroring `WithIdDocumentTextExtractionCategoryAttempts`, or (b) update the XML doc to state explicitly that subsequent calls replace the previous list. The params-style overload below is probably the friendliest public API:
  ```csharp
  public SdkConfigBuilder WithSuppressedScreen(string screen)
  {
      if (string.IsNullOrEmpty(screen))
          throw new ArgumentNullException(nameof(screen));
      if (_suppressedScreens == null)
          _suppressedScreens = new List<string>();
      _suppressedScreens.Add(screen);
      return this;
  }
  ```

**3. Missing tests for `SessionSpecification`/`SessionSpecificationBuilder` serialization**
- **File:** `test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs` (new tests added)
- **Issue:** The tests serialize `SdkConfig` in isolation. There's no end-to-end test that `SessionSpecification` (the outer DTO actually POSTed to the IDV backend) still produces the expected JSON when `suppressed_screens` is set. If anything in the outer envelope strips/renames the property in future, these tests won't catch it.
- **Suggested fix:** Add a test that builds a complete `SessionSpecification` via `SessionSpecificationBuilder().WithSdkConfig(...)` and asserts the serialized JSON includes `"suppressed_screens":[...]` in the expected location. If an existing `SessionSpecificationTests` file exists, extend it there to match the file-per-subject convention.

**4. `DeserializedSdkConfigShouldParseSuppressedScreens` depends on an unusual Json.NET behaviour that is worth an explicit test**
- **File:** `test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs:311-321`
- **Issue:** `SdkConfig` has only a 12-parameter constructor (no parameterless ctor) and all setters are get-only — deserialization relies on Json.NET matching the JSON key `suppressed_screens` to a constructor parameter (`suppressedScreens`) via its implicit constructor-matching. That works today because the names align, but it is fragile: a future refactor that renames the parameter (e.g. to `suppressed_screens_list`) would break deserialization without breaking the serialization test. There is no `[JsonConstructor]` attribute to pin this.
- **Suggested fix:** Either add `[JsonConstructor]` to the `SdkConfig` constructor and/or add an assertion to the test that round-trips *all* properties (not just `SuppressedScreens`) to lock in constructor-to-JSON name mapping. Alternatively, annotate each ctor parameter with `[JsonProperty("...")]` to make the mapping explicit.

### Nit

**1. XML documentation of valid constants should link by symbol, not free text**
- **File:** `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:244-249`
- **Suggested fix:** Replace the `<c>ID_DOCUMENT_EDUCATION</c>` etc. tokens with `<see cref="DocScanConstants.IdDocumentEducation"/>` so IDE/IntelliSense renders clickable links and renames stay in sync.

**2. Constant `FlowCompletion` ("FLOW_COMPLETION") is easy to confuse with the existing `SessionCompletion` ("SESSION_COMPLETION")**
- **File:** `src/Yoti.Auth/Constants/DocScanConstants.cs:36, 84`
- **Suggested fix:** If feasible, group the seven new suppressed-screen constants under a nested `public static class SuppressedScreens` inside `DocScanConstants` to scope them (`DocScanConstants.SuppressedScreens.FlowCompletion`). This also mirrors the `WithSuppressedScreens` setter name and makes IntelliSense discovery obvious.

**3. Missing trailing newline on modified files**
- **Files:** `src/Yoti.Auth/Constants/DocScanConstants.cs` (no newline at end of file, per diff), `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs`, `test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs`
- **Note:** Diff shows `\ No newline at end of file` on `DocScanConstants.cs`. Other files have been fixed. Keep consistent — add a trailing newline.

**4. BOM removed from `SdkConfig.cs`**
- **File:** `src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs:1`
- **Issue:** The diff shows the leading UTF-8 BOM (`﻿`) was stripped from this file while neighbouring files (e.g. `SdkConfigBuilderTests.cs`) still carry theirs. Minor codebase-consistency drift — no functional impact.
- **Suggested fix:** Re-save with the same BOM policy used across the solution (the repo appears to carry BOMs), or normalise the whole solution in a separate change.

**5. `WithSuppressedScreens` does not null-guard but lets `null` through**
- **File:** `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:253-257`
- **Note:** This is consistent with the rest of the builder (most setters accept anything and let the server validate). Just calling it out — no change required unless you want to be stricter.

---

### What's Done Well
- Test coverage is genuinely thorough for the size of the change: explicit positive, negative, serialization, deserialization, and forward-compatibility (unknown screen value) scenarios.
- `NullValueHandling = NullValueHandling.Ignore` on `suppressed_screens` is the correct choice — keeps the wire payload clean for callers who don't opt in.
- Constant values match the Yoti IDV shortened-flow public contract format (upper-snake-case), consistent with the rest of `DocScanConstants`.
- XML doc on `WithSuppressedScreens` enumerates the accepted values, which is a nice UX touch for SDK consumers.
- The test `SuppressedScreenConstantsShouldMatchExpectedValues` pins the wire values and will catch any accidental rename of a constant's value.

### Overall Score
**Approved with Comments** — safe to merge once the defensive-copy concern (Minor #1) and end-to-end `SessionSpecification` serialization test (Minor #3) are addressed; the rest are nice-to-haves.

PR #540, opened 2026-04-20

mehmet-yoti:
## Summary
This PR adds support for configuring the IDV shortened flow in the .NET SDK by introducing a `suppressed_screens` field to `SdkConfig`. Integrators can now specify which IDV UI screens to omit (e.g. education screens, flow completion screen) when creating a session, enabling a streamlined end-user journey.

## Changes
- **`DocScanConstants.cs`** — Added 7 new screen identifier constants (`SuppressedScreenIdDocumentEducation`, `SuppressedScreenIdDocumentRequirements`, `SuppressedScreenSupplementaryDocumentEducation`, `SuppressedScreenZoomLivenessEducation`, `SuppressedScreenStaticLivenessEducation`, `SuppressedScreenFaceCaptureEducation`, `SuppressedScreenFlowCompletion`) used to identify suppressible IDV screens.
- **`SdkConfig.cs`** — Added `SuppressedScreens` property (`List<string>`, JSON key `suppressed_screens`) and wired it through the constructor.
- **`SdkConfigBuilder.cs`** — Added `WithSuppressedScreens(IEnumerable<string>)` method with an allowlist validation guard; unknown screen identifiers throw `ArgumentException`. Added a static `AllowedSuppressedScreens` list that constrains valid inputs to the 7 defined constants.
- **`SdkConfigBuilderTests.cs`** — Added 6 new unit tests covering: null default, valid single/multiple/all screens, empty list, unknown screen rejection, and last-call-wins override behaviour.

## QA Test Steps

**Setup**
1. Build the solution: `dotnet build`
2. Run the test suite: `dotnet test`
3. Confirm all 6 new tests in `SdkConfigBuilderTests` pass alongside existing tests.

**Happy path — suppressing specific screens**
4. Construct an `SdkConfig` via `SdkConfigBuilder` calling `WithSuppressedScreens` with a list containing one or more of the valid constants (e.g. `DocScanConstants.SuppressedScreenFlowCompletion`, `DocScanConstants.SuppressedScreenIdDocumentEducation`).
5. Serialize the resulting `SdkConfig` to JSON (e.g. embed it in a `SessionSpecification`) and confirm the payload contains `"suppressed_screens": ["FLOW_COMPLETION", "ID_DOCUMENT_EDUCATION"]`.
6. Create an IDV session using the SDK against a test/sandbox environment and verify the specified screens are absent from the user-facing flow.

**Edge cases**
7. Call `WithSuppressedScreens` twice on the same builder and confirm only the second list is present in the built config (last-call-wins, `SuppressedScreensShouldOverridePreviousCallToWithSuppressedScreens`).
8. Call `WithSuppressedScreens` with an empty list — confirm `SuppressedScreens` is a non-null empty list (not null).
9. Do not call `WithSuppressedScreens` at all — confirm `SdkConfig.SuppressedScreens` is `null` and the JSON key is omitted from the serialized payload.
10. Call `WithSuppressedScreens` with an unrecognised string (e.g. `"UNKNOWN_SCREEN"`) — confirm an `ArgumentException` is thrown before `Build()` completes.
11. Pass all 7 valid screen constants simultaneously and confirm all are reflected in `SdkConfig.SuppressedScreens`.

**Regression**
12. Build an `SdkConfig` without calling `WithSuppressedScreens` and verify all pre-existing fields (`AllowedCaptureMethods`, `AttemptsConfiguration`, `AllowHandoff`, etc.) serialise correctly and are unaffected.

## Notes
- `WithSuppressedScreens` replaces the entire list on each call rather than appending; callers should pass the complete desired set in a single call.
- Validation is strict: only the 7 constants defined in `DocScanConstants` are accepted. Any future screens must be added to both `DocScanConstants` and the `AllowedSuppressedScreens` allowlist in `SdkConfigBuilder`.
- The `suppressed_screens` JSON field is omitted (not serialised as `null` or `[]`) when not set, preserving backwards compatibility with existing session payloads.

---

*Related Jira:* SDK-2612

### DIFF
--- a/src/Yoti.Auth/Constants/DocScanConstants.cs
+++ b/src/Yoti.Auth/Constants/DocScanConstants.cs
@@ -74,5 +74,13 @@
         public const string Generic = "GENERIC";
 
         public const string VerifyShareCodeTask = "VERIFY_SHARE_CODE_TASK";
+
+        public const string SuppressedScreenIdDocumentEducation = "ID_DOCUMENT_EDUCATION";
+        public const string SuppressedScreenIdDocumentRequirements = "ID_DOCUMENT_REQUIREMENTS";
+        public const string SuppressedScreenSupplementaryDocumentEducation = "SUPPLEMENTARY_DOCUMENT_EDUCATION";
+        public const string SuppressedScreenZoomLivenessEducation = "ZOOM_LIVENESS_EDUCATION";
+        public const string SuppressedScreenStaticLivenessEducation = "STATIC_LIVENESS_EDUCATION";
+        public const string SuppressedScreenFaceCaptureEducation = "FACE_CAPTURE_EDUCATION";
+        public const string SuppressedScreenFlowCompletion = "FLOW_COMPLETION";
     }
 }
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
@@ -41,6 +41,9 @@ namespace Yoti.Auth.DocScan.Session.Create
         [JsonProperty(PropertyName = "attempts_configuration")]
         public AttemptsConfiguration AttemptsConfiguration { get; }
 
+        [JsonProperty(PropertyName = "suppressed_screens")]
+        public List<string> SuppressedScreens { get; }
+
         public SdkConfig(string allowedCaptureMethods,
                             string primaryColour,
                             string secondaryColour,
@@ -51,7 +54,8 @@ namespace Yoti.Auth.DocScan.Session.Create
                             string errorUrl,
                             string privacyPolicyUrl,
                             bool? allowHandoff = null,
-                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null)
+                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null,
+                            List<string> suppressedScreens = null)
         {
             AllowedCaptureMethods = allowedCaptureMethods;
             PrimaryColour = primaryColour;
@@ -63,6 +67,7 @@ namespace Yoti.Auth.DocScan.Session.Create
             ErrorUrl = errorUrl;
             PrivacyPolicyUrl = privacyPolicyUrl;
             AllowHandoff = allowHandoff;
+            SuppressedScreens = suppressedScreens;
 
             if (idDocumentTextDataExtractionRetriesConfig != null)
             {
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using Yoti.Auth.Constants;
 
 namespace Yoti.Auth.DocScan.Session.Create
@@ -16,6 +17,7 @@ namespace Yoti.Auth.DocScan.Session.Create
         private string _privacyPolicyUrl;
         private bool? _allowHandoff;
         private Dictionary<string, int> _idDocumentTextDataExtractionAttemptsConfig;
+        private List<string> _suppressedScreens;
 
         /// <summary>
         /// Sets the allowed capture method to "CAMERA"
@@ -234,7 +236,36 @@ namespace Yoti.Auth.DocScan.Session.Create
         {
             WithIdDocumentTextExtractionCategoryAttempts(DocScanConstants.Generic, genericAttempts);
             return this;
-        }   
+        }
+
+        private static readonly IReadOnlyList<string> AllowedSuppressedScreens = new List<string>
+        {
+            DocScanConstants.SuppressedScreenIdDocumentEducation,
+            DocScanConstants.SuppressedScreenIdDocumentRequirements,
+            DocScanConstants.SuppressedScreenSupplementaryDocumentEducation,
+            DocScanConstants.SuppressedScreenZoomLivenessEducation,
+            DocScanConstants.SuppressedScreenStaticLivenessEducation,
+            DocScanConstants.SuppressedScreenFaceCaptureEducation,
+            DocScanConstants.SuppressedScreenFlowCompletion,
+        };
+
+        /// <summary>
+        /// Sets the list of screens to suppress in the IDV flow
+        /// </summary>
+        /// <param name="screens">Screen identifiers to suppress</param>
+        /// <returns>The <see cref="SdkConfigBuilder"/></returns>
+        /// <exception cref="System.ArgumentException">Thrown if any screen identifier is not a recognised value</exception>
+        public SdkConfigBuilder WithSuppressedScreens(IEnumerable<string> screens)
+        {
+            var screenList = screens?.ToList() ?? new List<string>();
+            var unknown = screenList.Except(AllowedSuppressedScreens).ToList();
+            if (unknown.Count > 0)
+                throw new System.ArgumentException(
+                    $"Unknown suppressed screen identifier(s): {string.Join(", ", unknown)}");
+
+            _suppressedScreens = screenList;
+            return this;
+        }
 
         /// <summary>
         /// Builds the <see cref="SdkConfig"/> based on values supplied to the builder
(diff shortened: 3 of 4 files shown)


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Summary
This PR adds `SuppressedScreens` configuration to `SdkConfig`/`SdkConfigBuilder` with seven screen identifier constants, a whitelist validation check in the builder, and six new unit tests. The serialization will work correctly because `DocScanService` uses `NullValueHandling.Ignore`, so an unset `SuppressedScreens` is omitted from the request body. Overall the feature is straightforward and follows the surrounding builder pattern, but there are a few correctness, API-design, and consistency concerns worth addressing.

---

### Critical
None.

---

### Major

**1. Validation runs in the setter, not at `Build()` — inconsistent with the builder pattern and couples error timing to the order of calls**
- File: `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:258-268`
- Issue: `WithSuppressedScreens` throws immediately when given an unknown value. Every other `With*` method in this builder (see `WithAllowedCaptureMethods`, `WithPrimaryColour`, `WithIdDocumentTextExtractionCategoryAttempts`, etc.) performs no validation — they are pure setters and `Build()` is a pure constructor. Mixing validation strategies within the same builder is confusing and creates an asymmetry for users.
- Why it matters: If a caller later adds a new constant to `DocScanConstants` but forgets to add it to the private `AllowedSuppressedScreens` list (easy to miss — they live in separate files), calls that should succeed will throw `ArgumentException`. Also, eager throwing prevents the common pattern of constructing a builder from partially-validated data and inspecting the resulting config.
- Recommendation: Either (a) remove the whitelist check entirely and let the backend reject unknown identifiers — this matches how `WithAllowedCaptureMethods` works with `Camera`/`CameraAndUpload` (no client-side validation) and makes the SDK forward-compatible with backend additions — or (b) if client-side validation is a product requirement, move it into `Build()` so behavior is deterministic regardless of call order, and make `AllowedSuppressedScreens` a single source of truth (see finding #2). Option (a) is the lower-risk, more consistent choice.

**2. Whitelist duplicates the constants list — easy to drift out of sync**
- File: `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:241-250` and `src/Yoti.Auth/Constants/DocScanConstants.cs:78-84`
- Issue: The seven `SuppressedScreen*` constants are declared in `DocScanConstants`, then listed again in `AllowedSuppressedScreens` inside the builder. Adding a new screen requires edits in two places with no compile-time link between them.
- Why it matters: A new constant added to `DocScanConstants` will silently be rejected at runtime by `WithSuppressedScreens` until someone remembers to update the builder's list.
- Recommendation: If the whitelist is kept, derive it reflectively, move the list onto `DocScanConstants` as the canonical source (e.g. `DocScanConstants.AllSuppressedScreens`), or drop the whitelist (see finding #1). A simpler alternative: expose strongly-typed helper methods (`WithSuppressedIdDocumentEducation()`, etc.) the same way `WithAllowsCamera()` / `WithAllowsCameraAndUpload()` wrap the generic `WithAllowedCaptureMethods` — this eliminates the string-validation problem altogether.

**3. Missing null/empty-element validation in `WithSuppressedScreens`**
- File: `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:258-268`
- Issue: A caller passing `new List<string> { null }` or `new List<string> { "" }` will bypass the whitelist with a non-helpful error — `Except` will include `null`/`""` in `unknown` and emit a confusing message like `Unknown suppressed screen identifier(s): `. Passing `null` for `screens` silently becomes "clear" semantics (`_suppressedScreens = new List<string>()`), which conflicts with the natural "not set" / "leave default" intent already covered by simply not calling the method.
- Why it matters: Silent null-coalescing changes user intent; users who pass `null` likely meant either "reset" or made a mistake — neither maps cleanly to "an empty list was configured" in the serialized payload.
- Recommendation: Use the project's existing `Validation.NotNull(screens, nameof(screens))` (see `src/Yoti.Auth/Share/ExtraData.cs:23` and `PolicyBuilder.cs:16` for the convention), and guard against null/empty string entries explicitly.

---

### Minor

**4. `.Except(...)` is O(n·m) on a `List<string>` — semantically wrong-typed**
- File: `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:261`
- Issue: `AllowedSuppressedScreens` is declared `IReadOnlyList<string>`. `Enumerable.Except` against a list lookup is linear per element. For seven entries this is irrelevant, but using a `HashSet<string>` is idiomatic for membership testing and better self-documents intent.
- Recommendation: If keeping the validation, declare the whitelist as `private static readonly HashSet<string> AllowedSuppressedScreens = new HashSet<string> { ... };`.

**5. Fully-qualified `System.ArgumentException` instead of using the imported namespace**
- File: `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:263`
- Issue: The file does not `using System;`, so `System.ArgumentException` is written out in full. Most other files in the project import `System` directly.
- Recommendation: Add `using System;` at the top and use `throw new ArgumentException(...)` for consistency.

**6. XML doc on `WithSuppressedScreens` builder method is sparse**
- File: `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs`
- Issue: The doc says "Sets the list of screens to suppress in the IDV flow" without enumerating the valid values or explaining what each screen means — consumers will have to read the source of `DocScanConstants` to find the set.
- Recommendation: On `WithSuppressedScreens`, list the accepted constants (or cross-reference `DocScanConstants.SuppressedScreen*`) in the `<remarks>` block, matching the richer XML doc pattern used by `WithAllowHandoff` and `WithIdDocumentTextExtractionReclassificationAttempts`.

**7. Missing test coverage — JSON serialization and null-argument behavior**
- File: `test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs`
- Issue: No test verifies that:
  - `SuppressedScreens` serializes to the JSON property `suppressed_screens` (the API contract depends on this).
  - An unset `SuppressedScreens` is omitted from JSON output (matters because `DocScanService.YotiDefaultJsonSettings` uses `NullValueHandling.Ignore`).
  - `WithSuppressedScreens(null)` behavior — whether by design or not, the current implementation silently creates an empty list.
  - Unknown identifiers mixed with valid ones still throw (current test only uses a single unknown value).
- Recommendation: Add a serialization round-trip test (follow the pattern in other `DocScan` test files), and add a test for `WithSuppressedScreens(null)` that pins down the intended behavior.

---

### Nit

**8. Field/constant ordering in the builder**
- File: `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:241-250`
- Issue: `AllowedSuppressedScreens` is declared in the middle of the file, between `WithIdDocumentTextExtractionGenericAttempts` and `WithSuppressedScreens`. Static readonly fields in this codebase are typically grouped with other fields at the top of the class.
- Recommendation: Move the static declaration up next to the other private fields (near line 20) for readability.

**9. Missing trailing newline in modified files**
- Files: `src/Yoti.Auth/Constants/DocScanConstants.cs`, `src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs`, `test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs`
- Issue: Git reports `\ No newline at end of file` for the edits.
- Recommendation: Add a trailing newline to each file.

**10. Constant naming: `SuppressedScreen*` prefix is verbose**
- File: `src/Yoti.Auth/Constants/DocScanConstants.cs:78-84`
- Issue: `DocScanConstants.SuppressedScreenIdDocumentEducation` reads awkwardly at call sites. A nested static class (e.g. `DocScanConstants.SuppressedScreens.IdDocumentEducation`) would be more idiomatic C# and shorter at the use site.
- Recommendation: Optional refactor — introduce `public static class SuppressedScreens { public const string IdDocumentEducation = "ID_DOCUMENT_EDUCATION"; ... }` inside `DocScanConstants`.

---

### What's Done Well
- The `SdkConfig` constructor change adds the new parameter with a default value of `null`, preserving backward compatibility.
- Test coverage is broad for the common scenarios: null/empty/full/override/invalid.
- The `suppressed_screens` JSON property name correctly follows the snake_case convention used by the rest of `SdkConfig`.
- Relying on `NullValueHandling.Ignore` means the new optional field won't leak into payloads of callers who don't set it.

---

### Overall Score
**Needs Revision** — primarily for the validation-strategy inconsistency (finding #1), the duplicated whitelist (finding #2), and the null-handling gap (finding #3). The rest are polish items.

PR #542, opened 2026-04-20
SDK-2616: .NET - Support configuration for IDV shortened flow - dotnet

mehmet-yoti:
## Summary
Adds support for configuring the IDV shortened flow by allowing relying parties to suppress specific screens in the Doc Scan / IDV session SDK config. A new `suppressed_screens` array is serialized on the `SdkConfig` payload, along with a `SuppressedScreen` constants class enumerating the valid values (`ID_DOCUMENT_EDUCATION`, `FLOW_COMPLETION`, etc.).

## Changes
- **`src/Yoti.Auth/DocScan/Session/Create/SuppressedScreen.cs`** (new): Constants class exposing the seven valid suppressible screen names.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs`**: Adds `SuppressedScreens` property (serialized as `suppressed_screens`) and a new optional constructor argument.
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs`**: Adds `WithSuppressedScreens(List<string>)` (replaces the list) and `WithSuppressedScreen(string)` (appends a single value, lazily initialising the list).
- **`src/Yoti.Auth/Constants/DocScanConstants.cs`**: Adds the same screen-name constants under the nested constants scope for parity.
- **`test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs`**: 12 new tests covering null default, single-value and list-based configuration, JSON round-tripping, empty arrays, and forward-compatible handling of unknown screen values.
- **`test/Yoti.Auth.Tests/DocScan/Session/Create/SessionSpecificationBuilderTests.cs`**: New test verifying that `suppressed_screens` is serialized through to the full `SessionSpecification`.

## QA Test Steps
1. **Setup**: Check out the branch, restore NuGet packages, and build the solution (`dotnet build`).
2. **Run unit tests**: `dotnet test test/Yoti.Auth.Tests/Yoti.Auth.Tests.csproj --filter "FullyQualifiedName~SdkConfigBuilderTests|FullyQualifiedName~SessionSpecificationBuilderTests"` — all tests should pass, including the 13 new `SuppressedScreens*` tests.
3. **Happy path – list setter**: Build an `SdkConfig` via `new SdkConfigBuilder().WithSuppressedScreens(new List<string>{ SuppressedScreen.IdDocumentEducation, SuppressedScreen.FlowCompletion }).Build();`. Serialize via `JsonConvert.SerializeObject` and confirm the JSON contains `"suppressed_screens":["ID_DOCUMENT_EDUCATION","FLOW_COMPLETION"]`.
4. **Happy path – single setter**: Call `.WithSuppressedScreen(SuppressedScreen.FlowCompletion).WithSuppressedScreen(SuppressedScreen.IdDocumentRequirements)` on a fresh builder and confirm both entries appear, in order, on the built `SdkConfig.SuppressedScreens`.
5. **Default behaviour**: Build an `SdkConfig` without calling either suppressed-screen method. Confirm `SuppressedScreens` is `null` and that the serialized JSON (with `NullValueHandling.Ignore`) does not contain a `suppressed_screens` key — ensures non-shortened-flow callers are unaffected.
6. **Empty list**: Pass `new List<string>()` to `WithSuppressedScreens` and confirm the resulting property is a non-null empty list, serialized as `"suppressed_screens":[]`.
7. **Replacement semantics**: Call `WithSuppressedScreens(listA)` and then `WithSuppressedScreens(listB)` on the same builder; confirm only `listB` is retained (list setter replaces, it does not merge).
8. **End-to-end session spec**: Build a full `SessionSpecification` via `SessionSpecificationBuilder().WithSdkConfig(...)`. Confirm the serialized spec JSON contains `suppressed_screens` nested under `sdk_config`.
9. **Regression – existing SdkConfig fields**: Run the full `SdkConfigBuilderTests` suite and confirm the pre-existing tests (colours, capture methods, attempts configuration, handoff, etc.) still pass — verifies the new optional constructor parameter didn’t break existing call sites.
10. **Integration (optional, against IDV backend)**: Create a session with `suppressed_screens` set to `["ID_DOCUMENT_EDUCATION","FLOW_COMPLETION"]` and confirm the IDV UI flow omits those screens as expected.

## Notes
- The `SuppressedScreen` values are duplicated in both `SuppressedScreen` (public, intended as the canonical user-facing constants) and `DocScanConstants` (for internal parity with other screen-name constants). Consider consolidating to a single source in a follow-up if desired.
- Validation of screen names is intentionally deferred to the backend — the client accepts arbitrary strings, so newly introduced suppressible screens can be passed without an SDK release (verified by the `SuppressedScreensShouldDeserializeUnknownValuesGracefully` test).
- `WithSuppressedScreens(null)` will clear any previously configured list and cause `SuppressedScreens` to be omitted from JSON; this is the documented “opt-out” behaviour.
- No changes to the request/response contract beyond the new optional `suppressed_screens` field — fully backwards compatible with existing RP integrations.

---

*Related Jira:* SDK-2616

### DIFF
--- a/src/Yoti.Auth/Constants/DocScanConstants.cs
+++ b/src/Yoti.Auth/Constants/DocScanConstants.cs
@@ -74,5 +74,13 @@
         public const string Generic = "GENERIC";
 
         public const string VerifyShareCodeTask = "VERIFY_SHARE_CODE_TASK";
+
+        public const string IdDocumentEducation = "ID_DOCUMENT_EDUCATION";
+        public const string IdDocumentRequirements = "ID_DOCUMENT_REQUIREMENTS";
+        public const string SupplementaryDocumentEducation = "SUPPLEMENTARY_DOCUMENT_EDUCATION";
+        public const string ZoomLivenessEducation = "ZOOM_LIVENESS_EDUCATION";
+        public const string StaticLivenessEducation = "STATIC_LIVENESS_EDUCATION";
+        public const string FaceCaptureEducation = "FACE_CAPTURE_EDUCATION";
+        public const string FlowCompletion = "FLOW_COMPLETION";
     }
 }
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs
@@ -1,4 +1,4 @@
-﻿using Newtonsoft.Json;
+using Newtonsoft.Json;
 using System.Collections.Generic;
 
 namespace Yoti.Auth.DocScan.Session.Create
@@ -41,6 +41,14 @@ namespace Yoti.Auth.DocScan.Session.Create
         [JsonProperty(PropertyName = "attempts_configuration")]
         public AttemptsConfiguration AttemptsConfiguration { get; }
 
+        /// <summary>
+        /// The screens to suppress from the IDV flow. Each value should be one of the
+        /// constants defined in <see cref="SuppressedScreen"/>. The field is optional —
+        /// when <c>null</c> or empty, no screens are suppressed.
+        /// </summary>
+        [JsonProperty(PropertyName = "suppressed_screens")]
+        public List<string> SuppressedScreens { get; }
+
         public SdkConfig(string allowedCaptureMethods,
                             string primaryColour,
                             string secondaryColour,
@@ -51,7 +59,8 @@ namespace Yoti.Auth.DocScan.Session.Create
                             string errorUrl,
                             string privacyPolicyUrl,
                             bool? allowHandoff = null,
-                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null)
+                            Dictionary<string, int> idDocumentTextDataExtractionRetriesConfig = null,
+                            List<string> suppressedScreens = null)
         {
             AllowedCaptureMethods = allowedCaptureMethods;
             PrimaryColour = primaryColour;
@@ -63,6 +72,7 @@ namespace Yoti.Auth.DocScan.Session.Create
             ErrorUrl = errorUrl;
             PrivacyPolicyUrl = privacyPolicyUrl;
             AllowHandoff = allowHandoff;
+            SuppressedScreens = suppressedScreens;
 
             if (idDocumentTextDataExtractionRetriesConfig != null)
             {
--- a/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
+++ b/src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs
@@ -16,6 +16,7 @@ namespace Yoti.Auth.DocScan.Session.Create
         private string _privacyPolicyUrl;
         private bool? _allowHandoff;
         private Dictionary<string, int> _idDocumentTextDataExtractionAttemptsConfig;
+        private List<string> _suppressedScreens;
 
         /// <summary>
         /// Sets the allowed capture method to "CAMERA"
@@ -234,7 +235,40 @@ namespace Yoti.Auth.DocScan.Session.Create
         {
             WithIdDocumentTextExtractionCategoryAttempts(DocScanConstants.Generic, genericAttempts);
             return this;
-        }   
+        }
+
+        /// <summary>
+        /// Sets the list of screens to suppress from the IDV flow. This allows the RP
+        /// to shorten the flow by omitting specific screens.
+        /// </summary>
+        /// <remarks>
+        ///     Use the constants defined in <see cref="SuppressedScreen"/> for valid values.
+        ///     Calling this method replaces any previously configured list.
+        /// </remarks>
+        /// <param name="suppressedScreens">The list of screens to suppress</param>
+        /// <returns>The <see cref="SdkConfigBuilder"/></returns>
+        public SdkConfigBuilder WithSuppressedScreens(List<string> suppressedScreens)
+        {
+            _suppressedScreens = suppressedScreens;
+            return this;
+        }
+
+        /// <summary>
+        /// Adds a single screen to the list of screens to suppress from the IDV flow.
+        /// </summary>
+        /// <remarks>
+        ///     Use the constants defined in <see cref="SuppressedScreen"/> for valid values.
+        /// </remarks>
+        /// <param name="suppressedScreen">The screen to suppress</param>
+        /// <returns>The <see cref="SdkConfigBuilder"/></returns>
+        public SdkConfigBuilder WithSuppressedScreen(string suppressedScreen)
+        {
+            if (_suppressedScreens == null)
+                _suppressedScreens = new List<string>();
+
+            _suppressedScreens.Add(suppressedScreen);
+            return this;
+        }
 
         /// <summary>
         /// Builds the <see cref="SdkConfig"/> based on values supplied to the builder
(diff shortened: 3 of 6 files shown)


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Critical
None

### Major
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:264-271` (`WithSuppressedScreen`) — missing argument validation and no de-duplication.** Comparable single-item adders in this codebase consistently validate the input and avoid duplicates. `NotificationConfigBuilder.WithTopic` (line 138-144) skips already-present values, and `RequestedWatchlistScreeningCheckBuilder.WithCategory` (line 41-47) calls `Validation.NotNullOrWhiteSpace` *and* dedupes. Without these, a caller can pass `null` and get a list whose serialized JSON contains `null` entries that the backend will likely reject, or accidentally suppress the same screen multiple times. Suggested fix:
  ```csharp
  public SdkConfigBuilder WithSuppressedScreen(string suppressedScreen)
  {
      Validation.NotNullOrWhiteSpace(suppressedScreen, nameof(suppressedScreen));
      if (_suppressedScreens == null)
          _suppressedScreens = new List<string>();
      if (!_suppressedScreens.Contains(suppressedScreen))
          _suppressedScreens.Add(suppressedScreen);
      return this;
  }
  ```

- **`src/Yoti.Auth/DocScan/Session/Create/SuppressedScreen.cs` and `src/Yoti.Auth/Constants/DocScanConstants.cs:78-84` — duplicated source of truth for the same constant strings.** The same seven `ID_DOCUMENT_EDUCATION` … `FLOW_COMPLETION` literals now live in two public types. If the spec ever renames one (or adds another), the two will drift silently and tests in `SuppressedScreenConstantsShouldMatchExpectedValues` will not catch it. Pick one home and have the other reference it (or remove the duplication entirely). The natural choice is to keep them in `SuppressedScreen` only (matches the `XxxScreen` "valid values" pattern used here) and drop the `DocScanConstants` additions, since nothing inside the SDK consumes the `DocScanConstants` versions.

### Minor
- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:250-254` (`WithSuppressedScreens`) — accepts `null` silently and overwrites accumulated state without flagging it.** Most other `With…(List<…>)` setters in the codebase (e.g. `OrthogonalRestrictionsFilterBuilder.WithIncludedCountries`) also accept arbitrary input, so this is consistent — but given the new `WithSuppressedScreen` adder accumulates state, calling `WithSuppressedScreen(...)` then `WithSuppressedScreens(null)` quietly throws away the accumulated list. Either document this in the XML remarks ("passing null clears the list") or `Validation.NotNull` the parameter.

- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:266-267` — single-line `if` without braces.** The rest of the file uses braces sparingly too (`WithIdDocumentTextExtractionCategoryAttempts` on line 176-182 also omits them), so this is internally consistent — but Yoti style elsewhere (`RequestedWatchlistScreeningCheckBuilder`) brackets multi-line bodies. Minor consistency call. Not blocking.

- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs:49-50` — `SuppressedScreens` exposed as mutable `List<string>`.** The class otherwise looks like an immutable DTO (all properties are `{ get; }`) but consumers can still mutate the list via `sdkConfig.SuppressedScreens.Add(...)` after construction. The same is technically true of `AttemptsConfiguration.IdDocumentTextDataExtraction`, so this matches the existing pattern, but a defensive copy in the constructor (or returning `IReadOnlyList<string>`) would be cleaner. Not blocking unless you want to tighten the surface.

- **`test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs:280-294` (`SuppressedScreensShouldNotSerializeWhenNull`) — the test imports `JsonSerializerSettings` directly rather than going through the SDK's serializer.** The production code uses `DocScanService.YotiDefaultJsonSettings` (private) for outgoing requests. The test asserts the framework respects `NullValueHandling.Ignore`, not that the SDK uses it. Consider exercising the same `SessionSpecification` round-trip via the public path, or at minimum a comment noting this is a contract test for the JSON settings the SDK itself applies.

- **Test coverage gaps:** No test verifies behaviour when `WithSuppressedScreen` is called after `WithSuppressedScreens(...)` has set a non-empty list (does it append or overwrite?). No test for the order-preservation contract that a `List<string>` implies. No negative test covering the `null` argument case for either method (see Major above). Suggest adding:
  ```csharp
  [TestMethod]
  public void WithSuppressedScreenShouldAppendToExistingList() { ... }
  ```

- **`src/Yoti.Auth/DocScan/Session/Create/SuppressedScreen.cs:5` — `cref` points to a specific overload signature.** `<see cref="SdkConfigBuilder.WithSuppressedScreens(System.Collections.Generic.List{string})"/>` will go stale if the parameter type ever changes (e.g. to `IEnumerable<string>`) and won't surface the sibling `WithSuppressedScreen` adder. Prefer the unqualified `<see cref="SdkConfigBuilder.WithSuppressedScreens"/>` reference and mention both methods in prose.

### Nit
- **`src/Yoti.Auth/Constants/DocScanConstants.cs:86` — file ends without a trailing newline** (`No newline at end of file` in the diff). `SdkConfig.cs` was fixed to add a trailing newline in this same change; do the same for `DocScanConstants.cs` for consistency.

- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs:1` — UTF-8 BOM removed.** Other files in the same directory (e.g. the test file at `SdkConfigBuilderTests.cs:1`) still carry a BOM. Trivial, but a mixed-BOM repo can confuse some tooling/diffs. Either keep or strip consistently across the touched files.

- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfigBuilder.cs:246` — XML doc says "Calling this method replaces any previously configured list."** This contract is worth surfacing in the `<summary>` (not just `<remarks>`) since it differs from how the analogous `WithTopic`/`WithCategory` accumulators work. Even better, name the two methods consistently with the existing pattern in this codebase: a singular `WithX(...)` that accumulates, and either no plural setter or a `ClearSuppressedScreens()` if a reset is needed.

- **`src/Yoti.Auth/DocScan/Session/Create/SdkConfig.cs:62-63` — adding a new optional ctor parameter is technically a binary-incompatible change** for any consumer that ILMerges or uses Mono/older runtimes (default-arg values are baked into call sites). Existing callers compiled against the old SDK will keep working at the source level but won't pick up the new default. Not actionable — just flagging if the SDK has a strict binary-compat policy. The `idDocumentTextDataExtractionRetriesConfig` parameter was added the same way historically, so precedent supports it.

- **`test/Yoti.Auth.Tests/DocScan/Session/Create/SdkConfigBuilderTests.cs:233-243` (`ShouldBuildWithAllKnownSuppressedScreens`) — asserts hard-coded literal values rather than using the constants** (e.g. `Assert.AreEqual("ID_DOCUMENT_EDUCATION", sdkConfig.SuppressedScreens[0]);`). This duplicates the `SuppressedScreenConstantsShouldMatchExpectedValues` test and means a constant rename would fail in two places. Consider asserting against `SuppressedScreen.IdDocumentEducation` here and keep the literal-value check in the dedicated constants test only.

